PR #28, opened 2023-09-21
Propose support for UTF-8 in metric and label names

ywwg:
We propose to add support for arbitrary UTF-8 characters in Prometheus metric and label names. While tsdb and the Prometheus code already support arbitrary UTF-8 strings, PromQL and the exposition format need a quoting syntax to make these values distinguishable.

Associated Prometheus issue: https://github.com/prometheus/prometheus/issues/12630

### DIFF
--- a/proposals/2023-08-21-utf8.md
+++ b/proposals/2023-08-21-utf8.md
@@ -1,0 +1,261 @@
+# UTF-8 Support for Metric and Label names
+
+* **Owners:**
+  * `<@author: owen.williams@grafana.com>`
+
+* **Implementation Status:** `Proof-of-concept implemented`
+
+* **Related Issues and PRs:**
+  * [GH Issue](https://github.com/prometheus/prometheus/issues/12630)
+  * [PR](https://github.com/grafana/mimir-prometheus/pull/476) (TODO: needs to be rebased on upstream prom)
+
+* **Other docs or links:**
+  * [Background Discussion / Justification](https://docs.google.com/document/d/1yFj5QSd1AgCYecZ9EJ8f2t4OgF2KBZgJYVde-uzVEtI/edit). Please read this document first for more information on the chosen solution.
+
+> TL;DR: We propose to add support for arbitrary UTF-8 characters in Prometheus metric and label names. While tsdb and the Prometheus code already support arbitrary UTF-8 strings, PromQL and the exposition format need a quoting syntax to make these values distinguishable.
+
+## Why
+
+We wish to use Prometheus as a backing database for other metrics implementations like Open Telemetry that use Prometheus-illegal characters, like '`.`'. Establishing a quoting syntax avoids the need for verbose and unreadable character encoding schemes, as well as non-roundtripable replacement schemes like using underscore for all unsupported characters.
+
+### Pitfalls of the current solution
+
+The current recommendation for users wanting to handle invalid characters is to replace all those characters with underscores. This substitution cannot be reversed, creating the possibility for name collisions. Furthermore, the change in metric names between original write and eventual read can be confusing for end users.
+
+## Goals
+
+* Allow arbitrarily-named metrics and labels to be submitted to Prometheus.
+* Allow as many UTF-8 characters as possible to exist literally inline, with escaping required for characters that are unprintable or cause parsing problems.
+* Allow these metric and label names to be easily queried back with PromQL.
+* Ensure that Go templating supports quoted label names.
+* Provide a graceful backwards-compatible fallback for query clients that do not support UTF-8 metric and label names.
+
+### Audience
+
+The audience for this change is very broad. It would allow all end-users of Prometheus to use richer metric names. Tool and service authors will want to be aware of the change and support the new character set.
+
+## Non-Goals
+
+We do not directly address the issue of adding function-like syntaxes, such as `metricname.sum`. The changes proposed here are forward-compatible with such a new syntax, as long as the new syntax is outside of the quoted metric names.
+
+We do not propose a syntax for selecting metric names other than by basic equality. The changes proposeed here are forward-compatible with other future possible selection mechanisms, for instance operator prefixes on the quoted metric name.
+
+## How
+
+### But First: Differences between PromQL and Openmetrics Exposition
+
+Changes will need to be made to both PromQL and the exposition format. It is important to note that while these two languages have surface similarities, there are important differences between them that need to be accounted for:
+
+* Quotes:
+  * **PromQL**: Strings may be quoted with single or double quotes. Strings may also be quoted with backticks to avoid escape sequences (raw strings).
+  * **Exposition format**: Supports only double quotes.
+* Equal Sign:
+  * **PromQL**: The equal sign (`=`) is a label matching operator, and there are other label matching operators for other operations.
+  * **Exposition Format**: The equal sign is an assignment operator that sets a label to a specific value.
+* String Escaping:
+  * **PromQL**: Strings are unquoted using the Go strconv.Unquote function.
+  * **Exposition Format**: Characters are escaped with a custom backslash format.
+
+### What will the new syntax look like?
+
+**tl;dr:**
+
+`{"my.noncompliant.metric", "noncompliant.label"="value"}`
+
+#### Oh no, I hate it
+
+A full discussion of rejected alternatives can be found below, but a common first reaction is, why move the metric name inside the braces? Why not leave it outside, like: `"my.noncompliant.metric"{"label"="value"}`.
+
+There are some specific risks with this alternative approach, but we also wanted to weigh the value of a syntax that is **more familiar to current experts** vs **more approachable for new users**. We take the optimistic position that there are more potential new users of Prometheus than there are existing users, and therefore decided to give more weight to that framing.
+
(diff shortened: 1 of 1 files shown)


bwplotka:
Thank you for this work!

We are discussing initial premises of this proposal in our DevSummit: https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#heading=h.f3063y1z43az

roidelapluie (approved):

roidelapluie:
LGTM

roidelapluie:
Did we consider instead of escaping for old clients, returning `{__name__="🎉"}` instead of `U___` ?

bwplotka:
> Did we consider instead of escaping for old clients, returning {__name__="🎉"} instead of U___ ?

This idea is great for metric names (U__ is not pretty I guess), but not for label names AFAIK? 🤔

proposals/2023-08-21-utf8.md
```diff
@@ -0,0 +1,261 @@
...
+There a few situations where we need to consider backwards compatibility:
+
+1. Old metrics provider, new backend: Old providers will not serve UTF-8 metric names, so nothing needs to be done here.
+2. New metrics provider, old backend: Old versions of Prometheus will throw an error if they see the new syntax. Therefore we must provide a way for the scraped target to serve previously-invalid metrics in a form that the old database can process.
+3. New backend, old query client: Errors will occur if an old client is fed UTF-8 metric names. Therefore we must provide a way for Prometheus to escape metric names such that old clients can query for and read them.
+4. Old backend, new query client: as with the first case, nothing needs to be done.
+
+### Text escaping
+
+We propose an escaping syntax for metric names that contain UTF-8 characters for systems that do not support UTF-8. This system can be used in queries and responses to provide compatibility between the systems.
```
bwplotka, line 133: When exactly this can/should be used?

proposals/2023-08-21-utf8.md
```diff
@@ -0,0 +1,261 @@
...
+* All non-valid characters (i.e. characters other than letters, numbers, and underscores) will be encoded as underscores surrounding the unicode value, like `_1F60A_`.
+* All pre-existing underscores will become doubled: `__`.
+* If a string should start with "U__" already, it will need to be escaped: `U___55_____`. (That's `U__` + `_55_` (for `U`) + `__` + `__`).
+
+#### Mixed-Block Scenario 
+
+We must consider an edge case in which a newer client persists metrics to disk in an older database that does not support UTF-8. Those metrics will be written to disk with the U__ escaping format. If, later, the user upgrades their database software, new metrics will be written with the native UTF-8 characters. At query time, there will be a problem: newer blocks were written with UTF-8 and older blocks were written with the escaping format. The query code will not know which encoding to look for.
+
+To avoid this confusion we propose to bump the version number in the tsdb meta.json file. On a per-block basis the query code can check the version number and look for UTF-8 metric names in native encoding for bumped version number 2, or in U__-escaped naming for earlier version 1.
```
bwplotka, line 144: 👍🏽

proposals/2023-08-21-utf8.md
```diff
@@ -0,0 +1,261 @@
...
+### Templating syntax
+
+Templating is used in alerts, and for dashboards that want to reference metric values. Go templating already supports quoting by way of the `index` keyword. Whereas labels are usually denoted as `{{ $labels.my_label_name }}`, the quote syntax is `{{ index $labels "my.label.name" }}`. No code changes are required.
+
+## Backwards-compatibility
+
+There a few situations where we need to consider backwards compatibility:
+
+1. Old metrics provider, new backend: Old providers will not serve UTF-8 metric names, so nothing needs to be done here.
+2. New metrics provider, old backend: Old versions of Prometheus will throw an error if they see the new syntax. Therefore we must provide a way for the scraped target to serve previously-invalid metrics in a form that the old database can process.
```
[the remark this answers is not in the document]
bwplotka: Ah you mention this in "### Scrape Time" - can you link that section in here?

proposals/2023-08-21-utf8.md
```diff
@@ -0,0 +1,261 @@
...
+
+To avoid this confusion we propose to bump the version number in the tsdb meta.json file. On a per-block basis the query code can check the version number and look for UTF-8 metric names in native encoding for bumped version number 2, or in U__-escaped naming for earlier version 1.
+
+Additionally, this means that an old metric or label name containing a valid escape sequence by chance would be parsed differently after enabling UTF-8 support. For example, a metric called `U___263a_` would appear as `☺` after the upgrade (if read from an old storage chunk, scraped from an old target, or received from an old remote-write client). This is an extremely unlikely scenario in practice so we believe a theoretically possible implicit name change is acceptable.
+
+Lastly, if a metric is written as "U__" but does not pass the unescaping algorithm, then we assume it was meant to be read as-is and let it pass through. 
+
+### Scrape Time
```
bwplotka, line 150: 👍🏽

bwplotka (approved):
Love it. Great job on proposal and figuring all (known to me so far) edge cases. 

LGTM 👍🏽

ywwg:
> Did we consider instead of escaping for old clients, returning `{__name__="🎉"}` instead of `U___` ?

the assumption I'm working on is that while the syntax may be parseable (excluding special chars like quotes and newlines), old code would not be expecting invalid metric names. Other code in the client may, for instance, perform a validity check, throw an error, or even crash if the metric name is not letters+numbers+_. Therefore I thought it was better to keep to the strict character set when returning results to old clients that do not advertise support for utf-8

proposals/2023-08-21-utf8.md
ywwg: I assume there are lots of home-brew applications that can query for PromQL, and I imagine some of them use an unofficial client library for a smaller language (ex: https://prometheus.io/docs/instrumenting/clientlibs/). The escaping system would prevent these applications from possibly breaking when they see utf8 metric names.

proposals/2023-08-21-utf8.md
```diff
@@ -0,0 +1,261 @@
...
+  * **Exposition Format**: The equal sign is an assignment operator that sets a label to a specific value.
+* String Escaping:
+  * **PromQL**: Strings are unquoted using the Go strconv.Unquote function.
+  * **Exposition Format**: Characters are escaped with a custom backslash format.
+
+### What will the new syntax look like?
+
+**tl;dr:**
+
+`{"my.noncompliant.metric", "noncompliant.label"="value"}`
```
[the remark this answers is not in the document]
bwplotka: Thanks!

ywwg:
Can someone approve the workflow to finish the automated checks?

ywwg:
@bwplotka do you think we're good to merge?

bwplotka (approved):
Sorry for lag. I changed settings so you should have checks without approve.

I will send reminder to Prometheus team as it's very large change that changes PromQL a lot, but so far no one gave any objection (: I will set deadline to the Wed next week initially for objections.

Thanks!

proposals/2023-08-21-utf8.md
bwplotka: 👍🏽

jesusvazquez (commented):
Nice work @ywwg writting this up and @bwplotka and @roidelapluie with the reviews. I'm happy with the end result and don't have much feedback to add.

After reading the doc and the backwards-compatibility section I'm still not certain about the kind of breaking changes this may create if any. Could you clarify if you are expecting any? Judging from the action plan it seems we should be safe with the feature flag but I wonder if any internal structure needs to be modified to accommodate the new characters (hopefully we are covered by the fact that we were already using string variables)

ywwg:
Bryan said that the [changes in the exposition format](https://github.com/ywwg/proposals/blob/owilliams/utf8-squash/proposals/2023-08-21-utf8.md#scrape-time), which require bumping the http version numbers, could be considered breaking changes. So, it depends on what the definition of "breaking" is, but as written, the new system should be backwards-compatible so I don't see any ABI/API breaks per se.

roidelapluie:
it is breaking for the client libraries, prometheus will still scrape 0.4

beorn7:
> After reading the doc and the backwards-compatibility section I'm still not certain about the kind of breaking changes this may create if any. Could you clarify if you are expecting any?

Probably the other comments already explained this, but to phrase it in my own words: If an instrumented target just started to expose in the new style, i.e. `{"the metric name", "björn"="rabenstein"}`, an old-style scraper would choke on it. That's why the "new style" has to be negotiated somehow, and the default behavior has to be to keep serving "old style", with the escaping schema described in the doc.

Similar is true for remote write.

> I wonder if any internal structure needs to be modified to accommodate the new characters

I don't think so. TSDB internally is not making any assumptions beyond "UTF-8 strings". We might have overlooked some things, but that would be an easy fix.

bwplotka:
Hm, I think what we discussed here is explained in [the proposal](https://github.com/prometheus/proposals/pull/28/files#diff-2844ee415793207b593c1a4231d67ab75a216cee6f153429c0547dfe5c5b9785R156), no?

> In the case of text scraping, we propose a version bump in the [TextVersion and OpenMetricsVersion numbers](https://github.com/prometheus/common/blob/main/expfmt/expfmt.go) so that a Prometheus scraper can signal its support of the new syntaxes. For TextVersion, this would be version 0.0.5. For OpenMetrics this would either be version 1.1.0 or 2.0.0, depending on whether this is considered a breaking change.

So yea, we have to bump protocol version to start sending UTF-8, and if someone sends UTF-8 with 0.0.4 that's violated protocol. That's not breaking change, that's just requirement to use a new future (clients to adopt new protocol version).

jesusvazquez (commented):
No more questions from me :wink: Lets merge :rocket:

bwplotka:
💪🏽 Thanks All!

PR #29, opened 2023-10-03
Add proposal for created timestamps

ArthurSens:
After approval in the [last dev summit](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#heading=h.yfs9vylmqi6r), I'm moving the proposal from the [design doc](https://docs.google.com/document/d/1kakDVn8aP1JerimLeazuMfy2jaB14W2kZDHMRokPcv4/edit#heading=h.cmqufn5ktjuv) to here :)


proposals/2023-06-13_created-timestamp.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Alternatives
+
+### Inject sentinel values
+
+Instead of injecting zero samples, a sentinel value, e.g. _CreatedNaN_ could be ingested intead to avoid confusion when noticing that not all samples follow the configured scrape interval.
+
+It would also help remote-write ingestors to easily identify created timestamps in a batch of samples.
+
+It would require refactoring of rate-like functions, where sentinel values would need to be replaced with zeros.
```
MichaHoffmann, line 132: I think zero injection might mess with "count_over_time" and other horizontal aggregations maybe, so probably this also needs to be addressed in the query layer right?

beorn7: Yes, it will change the outcome of `count_over_time`, that's why it has to be behind a feature flag.

The question is more if it's a change for the better or for the worse.

ArthurSens: It's a good point though, it should be mentioned in the proposal

bwplotka:
Thanks! The only chance I would see is to mention the plan: Adding zero injection and then metadata long term - as agreed in DevSummit (:

proposals/2023-06-13_created-timestamp.md
ArthurSens: added this problem to the "Feature flag" section

ArthurSens:
> Thanks! The only change I would propose is to mention the plan: Adding zero injection and then metadata long term - as agreed in DevSummit (: 

Have you seen the "Action plan" section? Were you thinking of something different?

> Let's also mention feature flag.
Section added!

proposals/2023-06-13_created-timestamp.md
```diff
@@ -0,0 +1,158 @@
...
+On following scrapes, created timestamps are not appended until a scrape finds a created timestamp higher than the latest sample's timestamp, which would trigger the insertion of another sample with a 0-value sample.
+
+![Alt text](../assets/2023-06-13_created-timestamps/followup-scrape-example.png)
+
+This is already enough to improve rate calculations as mentioned in the [Goals](#goals) section, while also persisting such data in WAL (even if not explicit) and although it allows an easier implementation, it comes with few limitations:
+
+* Unable to ingest created timestamps out of TSDB Head range.
+* Unable to help ingestors that require created timestamps as a separate field, as they would need to implement stateful logic to identify injected zeros in a batch of samples.
+
+For those reasons, created timestamps will also be stored as metadata **per series**, following the similar logic used for the zero-injection.
```
metalmatze, line 76: It is not entirely clear from this sentence why the synthetic samples are still needed if the created timestamps are also stored as metadata. 
Am I reading this correctly that, the created timestamps are therefore stored twice?

proposals/2023-06-13_created-timestamp.md
```diff
@@ -0,0 +1,158 @@
...
+
+#### Prometheus Protobuf
+
+Since Prometheus already supports scraping endpoints with Prometheus protobuf and knowing that adding the created timestamp field is trivial, it makes Prometheus protobuf the preferred option for initial implementation:
+
+```proto
+message Counter {
+  optional double   value    = 1;
+  optional Exemplar exemplar = 2;
+  optional google.protobuf.Timestamp    created_timestamp_ms = 3;
```
metalmatze, line 90:
```suggestion
  optional google.protobuf.Timestamp    created_timestamp = 3;
```

Does it really make sense to have the `_ms` suffix here, when the type is a `protobuf.Timestamp` that internally has seconds and nanos? I'd say we either remove the suffix or using a int64 and call it as proposed.

metalmatze: One could even argue that timestamp is also obsolete given the type is of timestamp. So potentially all we would need is: 
```proto
optional google.protobuf.Timestamp    created = 3;
```

proposals/2023-06-13_created-timestamp.md
ArthurSens: Yes, they are stored twice

* Injected in TSDB head so we can better rate results without changing interfaces to make metadata available during query time. Also, it would probably have a performance impact if we needed to inject zeros for every query.
* Added to metadata again, to enable remote-write receivers to easily identify created timestamps.

proposals/2023-06-13_created-timestamp.md
ArthurSens: Good point about the ms suffix, definitely not needed and in fact it was already merged downstream 😅 
https://github.com/prometheus/client_model/pull/66

I'm fixing the proposal

ArthurSens: Regarding also removing the timestamp suffix, I think we could keep it because all discussions around the subject we keep mentioning "created timestamp" instead of just "created". It makes things easier to understand

metalmatze: Sounds good to me.

proposals/2023-06-13_created-timestamp.md
metalmatze: Thanks, the first part was the explanation I was looking for. Thanks for writing it!

bwplotka (approved):
Epic work @ArthurSens thanks 💪🏽

PR #30, opened 2023-11-15
Proposal Amendment: UTF-8 migration

ywwg:
This proposal expands upon the existing UTF-8 proposal, providing more detail into how we plan to handle the migration of metrics data, ensuring that it remains queryable during the transition period


ywwg:
TODO: need to clarify expansion method, otherwise first draft ready to go

beorn7:
I added @roidelapluie and @bwplotka as reviewers of the original UTF-8 design doc, and @gouthamve as he was involved discussing the approach.

bwplotka (commented):
Nice, this is quite a smart way to have better experience on those metrics, especially given past decision on yolo underscores. Great work! I think it will massively help on UX with already ingested Otel metrics 💪🏽 

The collisions are tricky case here indeed. Hard to say if there will be case that might surprise users. We don't have specific examples that semantically makes sense. Given we cannot rule out the surprises completely, I would vote for dropping that storage trick. It's quite a complexity, yet still prone to surprise situations. Additionally, if we have block with mixed client versions (prone to collisions) and it's get compacted into 2 days or 2w block (e.g. Thanos), that means whole 2w is prone to collisions right? Or we change compaction code too?

I would say let's make unsupported input for this migration feature explicitly unsupported, not "sometimes" unsupported. I wonder if we can simply rely on the query API parameter (and feature flag) to allow some debuggability/quick opt-out, without storage deduction and be explicit in docs etc

Which brings an interesting question, would this feature be part of an official PromQL spec? (aka require every PromQL exposed backend to have "UTF-8 compatibility broad" lookup, to be compatible) or is it an optional feature?

Having explicit regex makes sense.

Quick suggestions:
* Put storage trick to alternatives as some future improvement if we know about specific collision case which makes sense.
* Propose query API param (?)
* Clarify if this is part of the official PromQL spec or just a migration friendly Prometheus feature

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,151 @@
...
+
+Given a query for a UTF-8 metric or label name, the tsdb will look for that name in on-disk blocks whether those blocks were written in native UTF-8 or either of two supported name-munging patterns.
+Those series will be located even in cases when a single block has one metric written in more than one way.
+The tsdb will know what versions of clients wrote those blocks based on a new entry in the meta.json.
+
+### Mixed-Format Scenarios
+
+We must consider edge cases in which a blocks database has persisted metrics or labels that may have been written by different client versions. There are multiple ways this can (and will) happen:
+
+* A newer client persists names to an older database version. In this case, names would be escaped with the U__ syntax.  If the database is upgraded, newer blocks will be written in UTF-8.
```
gouthamve, line 44: What do you mean by database upgrade? Do you mean a new Prometheus version?

gouthamve (commented):
Thinking more, I think the implementation could get complicated.

Right now, the TSDB interface has no concept of "ingestion-version", and I would be uncomfortable plumbing it through for just this use-case. 

Instead could we do the following:
```
-promql.utf8_migration.enabled=true
-promql.utf8_migration.until=<date-time> (optional)
```

The migration would then become:
1. Set the first flag
2. Move everything over to UTF-8
3. Wait a bit
4. Set the second flag
5. Once retention is passed, unset the first flag

This might work tbh. The challenge is likely to be that step 2. will take a long time as most folks don't have control over the clients.

proposals/2023-11-13-utf8-migration.md
ywwg: yes that's what I meant here, I can fix the language

ywwg:
> -promql.utf8_migration.enabled=true

is the idea that 100% of queries would get replacement-expansion, rather than trying to figure out which blocks are which?

ywwg:
or does the date-time determine for what periods the expansion will take place?

bwplotka:
Both (: 

If the `-promql.utf8_migration.until=<date-time>` is not provided, we query all blocks with extended lookup. If it's provided we perform extended lookup only until given time.

gouthamve:
Yup, what Bartek explained.

proposals/2023-11-13-utf8-migration.md
ywwg: fixed

ywwg:
Addressed notes

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Proposed Solution
+
+For queries to return correct data we must differentiate the three cases above, and to do that we first propose to bump the version number in the tsdb meta.json file.
+On a per-block basis, the query code can check the version number and know if the data was written with an old version of the Prometheus code.
+This helps distinguish the first case.
+
+Secondly we will add two new flags to help define the range of dates that are affected by mixed blocks and will be used to distinguish the second case from the third.
+
+* `-promql.utf8_migration.enabled`: This flag tells Prometheus that a migration is in progress, meaning that blocks with the newer tsdb version number may be mixed. If this flag is false, blocks with the newer version number are understood to not be mixed and are exclusively UTF-8.
```
bwplotka, line 71: Nice!

Some suggestions:

1. Hm can we simplify the description? I am not sure if this feature has to block aware? It's essentially enabling PromQL engine to repeat metric lookups for different escaping techniques, depending on timeline, nothing else?

2. Also we could be more explicit with the flag name. Perhaps "promql.utf8_broad_lookup.(...)" would be more explicit?

3. Finally, I wonder if the next feature user would ask is to give option for what escaping techniques to lookup for. For many users e.g. not using DataDog it's clear we don't need to spend cycles on that escaping. To me this means boolean flag could be replaced with string array with some enum like`promql.utf8_broad_lookup.escape_formats`= "_", "U__", "DataDog" 🤔

```suggestion
* `-promql.utf8_broad_lookup.escape_formats`: This flag tells PromQL engine what escaping methods might have been previously used to escape UTF-8 characters. This is then used to transparently repeat series lookups for metric names or label names when UTF-8 characters are spotted, for each escaping format. Available values: ...
```

bwplotka (commented):
Epic, thanks! Final suggestions 🤗 (I hope), otherwise LGTM

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,148 @@
...
+
+All of these situations can be summarized as follows:
+
+1. Data written with old database code: all names are guaranteed not to be UTF-8.
+2. Data written with new database code by new clients: all names are guaranteed to be UTF-8-compatible.
+3. Data written with new database code by one or more old clients (and possibly new clients as well): No guarantees, some names could be escaped, others not.
+
+### Proposed Solution
+
+To help alleviate this confusion we first propose to bump the version number in the tsdb meta.json file. On a per-block basis, the query code can check the version number and know if the data was written with an old version of the database code. This helps distinguish the first case.
```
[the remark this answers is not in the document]
ywwg: The version number bump indicates if the block was written by utf-capable prometheus, but it does not indicate if the clients sent utf8 metrics or pre-escaped metrics (or if the block contains a combination of both).

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+Given a query for a UTF-8 metric or label name, the tsdb will look for that name in on-disk blocks whether those blocks were written in native UTF-8 or either of two supported name-escaping patterns.
+Those series will be located even in cases when a single block has one metric written in more than one way.
+The tsdb will differentiate those blocks based on entries in the meta.json and a new flag.
+
+### Mixed-Format Scenarios
+
+We must consider edge cases in which a blocks database has persisted metrics or labels that may have been written by different client versions. There are multiple ways this can (and will) happen:
+
+* A newer client persists names to an older Prometheus version. In this case, names would be escaped with the U__ syntax.  If Prometheus is upgraded, newer blocks will be written in UTF-8.
```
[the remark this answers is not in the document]
ywwg: Yes we assume that they will do A. B seems unlikely... more likely they will just keep sending _ or `_dot_`. But I guess we should cover case B just in case.

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Time Scope
+
+The issue of mixed-format blocks will persist for the retention period of the tsdb.
+For some deployments this means only 14 days, for others it may be on the order of years of persisted old data.
+
+### Proposed Solution
+
+For queries to return correct data we must differentiate the three cases above, and to do that we first propose to bump the version number in the tsdb meta.json file.
+On a per-block basis, the query code can check the version number and know if the data was written with an old version of the Prometheus code.
```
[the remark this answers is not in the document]
ywwg: Yes I suppose the only real benefit is removing the UTF-8 query from the list of possibilities. The other escapings may all be possible. Do you think that maybe we don't need a version number bump at all and can just use the flag/date-based logic?

proposals/2023-11-13-utf8-migration.md
ywwg: I agree that essentially the logic is date-based, not block based, so yes that does get simpler. (Unless we keep the tsdb version bump.... but that does seem to add complexity).

Let's start with simple and see how the performance is

ywwg:
had to fix some commit signing, hopefully I didn't break anything

proposals/2023-11-13-utf8-migration.md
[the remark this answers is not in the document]
bwplotka: Yea, I am fine with both.

It feels like small change to bump version and might unlock some optimizations, so I am fine keeping this up, just wanted to clarify.

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+This helps distinguish the first case.
+
+Secondly we will add two new flags to help define the range of dates that are affected by mixed blocks and will be used to distinguish the second case from the third.
+
+* `-promql.utf8_broad_lookup.escape_formats`: This flag tells PromQL engine what escaping methods might have been previously used to escape UTF-8 characters. This is then used to transparently repeat series lookups for metric names or label names when UTF-8 characters are spotted, for each escaping format. Available values will be a short enum representing underscores, U__, or dots-only escaping.
+* `-promql.utf8_migration.until=<date-time>`: This flag indicates the latest date-time (inclusive) for blocks that may contain mixed data. Any data after this moment are exclusively UTF-8.
```
bwplotka, line 72:
```suggestion
* `-promql.utf8_broad_lookup.until=<date-time>`: This flag indicates the latest date-time (inclusive) for blocks that may contain mixed data. Any data after this moment are exclusively UTF-8.
```

Perhaps?

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Migration Timeline
+
+A Prometheus migration to UTF-8 will follow this timeline:
+
+1. Prometheus is upgraded and UTF-8 support enabled. The `-promql.utf8_broad_lookup.escape_formats` is turned on immediately, enabling the multi-lookup behavior and listing the possible escaping schemes.
+2. Clients are gradually upgraded to UTF-8.
+3. `-promql.utf8_migration.until` is set to the last date-time when a non-UTF-8 client sent data.
```
bwplotka, line 80:
```suggestion
3. `-promql. utf8_broad_lookup.until` is set to the last date-time when a non-UTF-8 client sent data.
```

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+4. Wait for the retention period to elapse such that the migration-until date is expired (could be years).
+5. The migration is complete. Remove `-promql.utf8_broad_lookup.escape_formats` and `-promql.utf8_migration.until` as they are no longer needed.
```
bwplotka, line 82:
```suggestion
5. The migration is complete. Remove `-promql.utf8_broad_lookup.escape_formats` and `-promql. utf8_broad_lookup.until` as they are no longer needed.
```

bwplotka (approved):
LGTM, thanks! 💪🏽

bwplotka:
> Which brings an interesting question, would this feature be part of an official PromQL spec? (aka require every PromQL exposed backend to have "UTF-8 compatibility broad" lookup, to be compatible) or is it an optional feature?

I assume the answer is NO, it's optional backend feature and not part of the PromQL spec (:

proposals/2023-11-13-utf8-migration.md
ywwg: applying manually

ywwg:
> I assume the answer is NO, it's optional backend feature and not part of the PromQL spec (:

I agree, I think this is an optional feature and not a requirement.

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+### Proposed Solution
+
+For queries to return correct data we must differentiate the three cases above, and to do that we first propose to bump the version number in the tsdb meta.json file.
+On a per-block basis, the query code can check the version number and know if the data was written with an old version of the Prometheus code.
+This helps distinguish the first case.
+
+Secondly we will add two new flags to help define the range of dates that are affected by mixed blocks and will be used to distinguish the second case from the third.
+
+* `-promql.utf8_broad_lookup.escape_formats`: This flag tells PromQL engine what escaping methods might have been previously used to escape UTF-8 characters. This is then used to transparently repeat series lookups for metric names or label names when UTF-8 characters are spotted, for each escaping format. Available values will be a short enum representing underscores, U__, or dots-only escaping.
+* `-promql.utf8_broad_lookup.until=<date-time>`: This flag indicates the latest date-time (inclusive) for blocks that may contain mixed data. Any data after this moment are exclusively UTF-8.
```
beorn7, line 72: Maybe describe the default case? I guess the default ist "forever", but the flag only becomes effective once at least one escape format has been specified with the other flag.

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+### Querying Mixed Blocks
+
+The last major challenge is correctly returning data for queries of blocks that contain mixed data.
+For the mixed-format scenarios, at query time, we will look for **all possible** escapings of a name in order to locate the correct data.
+We propose to do this by expanding a lookup for a UTF-8 metric or label name into a limited set of possible escapings:
+
+1. **UTF-8**
+2. **underscore-replaced**: All unsupported characters are converted to underscores.
+3. **U__ escaping**:  As described in the UTF-8 proposal, strings with invalid characters can be escaped by prepending `U__` and replacing all invalid characters with `_[UTF8 value]_`.
+4. **[Datadog proxy](https://github.com/grafana/mimir-proxies/blob/main/pkg/datadog/ddprom/naming.go#L30-L34) escaping pattern**: "`.`" becomes "`_dot_`" and "`_`" becomes "`__`".
```
beorn7, line 93: The flag `-promql.utf8_broad_lookup.escape_formats` implies that the search is constrained to the escaping schemas given by the flags. This is mentioned below (in line 108). How about mentioning this up here, so that the reader will immediately understand that we don't have to brute-force all methods if only a few have been specified?

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+* A newer Prometheus receives names from an older client, which is later upgraded. In this case, older names might be escaped using the replace-with-underscores method, and newer names will be UTF-8. This will often happen when Prometheus is receiving Open Telemetry metrics.
+* A newer Prometheus receives names from a mix of new and old clients, in which case the same block could contain escaped and UTF-8 data representing the same intended names.
+
+At query time, there will be a problem: some data may be written with UTF-8 and other data was written with an escaping format.
+The query code will not know which encoding to look for.
+In order to ensure consistent querying, the backwards-compatibility design must account for these scenarios, making trade-offs when needed.
+
+All of these situations can be summarized as follows:
+
+1. **Old Data** -- Data written with old Prometheus code: all names are guaranteed not to be UTF-8.
```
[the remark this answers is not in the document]
beorn7: After having read through all the below, I would say what's meant here is "if at all, we will have escaped names, so we would never query for UTF-8 names, but only try out the specified escaping schemas".

Still, I think the difference to the mixed data case should be made clearer. 

In principle, there might be a scenario where the user knows for sure that they won't have any escaping at all, for example if they had a pure Prometheus stack so far (no OTel etc.), but they would like to use UTF-8 names, so they still have a period of mixed versions deployed (new and old producers, new and old ingesters). I guess it's fine to not implement a specific optimization for that use case (which would be that we don't need a broad search for the old blocks at all), but it would be good if that case is described and that we disregard it as an informed decision.

proposals/2023-11-13-utf8-migration.md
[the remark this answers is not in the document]
beorn7: So yeah, I guess the only optimization is that we never have to run the original UTF-8 query on blocks with an older version. Not sure if that is worth the effort of versioning the blocks. But maybe, as @bwplotka said, the effort is actually quite low, and we might see additional uses of the version number later.

proposals/2023-11-13-utf8-migration.md
```diff
@@ -0,0 +1,174 @@
...
+### Lookup Table / Per-Name Config
+
+We considered recording a lookup table or per-name configuration that would describe how UTF-8 metrics and labels might be stored in old data blocks.
+This approach would be faster than doing query expansion, but would create extra operational overhead -- lookup tables would have to be correct and exhaustive.
+
+Because names are stored in the index, query expansion is not expensive enough to justify the extra operational overhead.
+
+### No Migration -- Write Both Versions
+
+We very briefly considered the idea of having the tsdb write all names for a name as long as the user configured it that way.
```
beorn7, line 169: names for a name → versions of a name

beorn7 (changes requested):
Thank you very much. I think this is substantially good to go. My comments are mostly about wordsmithing and clarifications.

ywwg:
all notes addressed, please let me know if I missed anything

beorn7 (approved):
LGTM. I assume all other reviewers are also happy at this point. Please follow up if not.

PR #31, opened 2024-01-26
Store and query classic histograms as native histograms with custom buckets

krajorama:
To support [RW Self-Contained Histograms](https://docs.google.com/document/d/1mpcSWH1B82q-BtJza-eJ8xMLlKt6EJ9oFGH325vtY1Q/edit?pli=1#heading=h.ueg7q07wymku) which is about the need to make writing histograms atomic, in particular to avoid a situation when series of a classic histogram are partially written to (remote) storage. For more information consult the referenced design document.

To make storing classic histograms more efficient by taking advantage of the design of native histograms.

Finally, fully custom bucket layouts is a larger project with wider scope. By reducing the scope we can have a shorter development cycle and offer a good feature and savings sooner.

### DIFF
--- a/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
+++ b/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
@@ -54,7 +54,7 @@ Finally, fully custom bucket layouts is a larger project with wider scope. By re
 ### Naming convention
 
 For the sake of brevity will use the following wording:
-* Classic histogram = classic histogram implying separate series for buckets, count and sum.
+* Classic histogram = Original Prometheus histogram implemented by separate series of buckets metrics labeled by `le`, with separate `_count` and `_sum` metrics.
 * Native histogram = a histogram that uses a single series to store all relevant information in complex data type samples.
 * Exponential histogram = a native histogram with exponential buckets.
 * Custom histogram = a native histogram with custom (i.e. user defined) buckets.
--- a/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
+++ b/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
@@ -87,6 +87,8 @@ Scenarios provided that the native histograms feature is enabled:
   * If the scrape histogram has exponential buckets, no effect. Exponential buckets have higher priority.
   * Otherwise the new custom histograms are scraped in a series without suffix.
 
+> Note: enabling native histograms comes with all notes and warnings described in the documentation, in particular the impact on the `le` label for the classic histograms for certain clients. See "Note about the format of le and quantile label values" in [Native Histograms](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms).
+
 ### Scrape
 
 * Create the new representation in a series named `<metric>` without the `le` label. The resulting series name would be the same as if the user enabled exponential histograms. For example `http_request_latency_seconds_bucket`, `http_request_latency_seconds_count` and `http_request_latency_seconds_sum` will simply be stored as a one metric called `http_request_latency_seconds` from now on.
@@ -137,6 +139,12 @@ Use case: custom histograms feature is off, but exponential histograms (current 
 
 * For histograms that have no exponential histogram version, the same applies as above.
 * For histograms that have exponential histogram version, nothing happens, exponential histograms will be kept.
+
+### Documentation
+
+* Feature should be marked as experimental, it depends on another experimental feature, namely native histograms.
+* Configuration option to be documented.
+* Impact on queries and PromQL functions to be specified.
 
 ### Data structures
 
--- a/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
+++ b/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
@@ -75,6 +75,18 @@ In this option only instrumentation backwards compatibility is guaranteed(*). Th
 
 *: there is still the option to keep scraping classic histograms series beside custom histograms to be able to use the old queries, in fact it’s a recommended migration tactic, see later.
 
+### Scrape configuration
+
+1. The feature is disabled if the feature flag [`native-histograms`](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms) is disabled.
+2. If native histograms feature is enabled, custom histograms can be enabled in `scrape_config` by setting the configration option `scrape_custom_histograms` to `true`.
+
+Scenarios provided that the native histograms feature is enabled:
+* If the scraped metric has custom bucket definitions and `scrape_classic_histograms` is enabled, the original classic histogram series with suffixes shall be generated. This is independent of the setting for custom histograms.
+* If `scrape_custom_histograms` is disabled, no effect.
+* If `scrape_custom_histograms` is enabled:
+  * If the scrape histogram has exponential buckets, no effect. Exponential buckets have higher priority.
+  * Otherwise the new custom histograms are scraped in a series without suffix.
+
 ### Scrape
 
 * Create the new representation in a series named `<metric>` without the `le` label. The resulting series name would be the same as if the user enabled exponential histograms. For example `http_request_latency_seconds_bucket`, `http_request_latency_seconds_count` and `http_request_latency_seconds_sum` will simply be stored as a one metric called `http_request_latency_seconds` from now on.
@@ -84,6 +96,10 @@ In this option only instrumentation backwards compatibility is guaranteed(*). Th
   * In the Prometheus text exposition format values are represented as floats. The dot (`.`) is not mandatory.
   * In the OpenMetrics text exposition [format](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#numbers) floats and integers are explicitly distinguished by the dot (`.`) sign being present or not in the value.
   * In the Prometheus/OpenMetrics ProtoBuf [format](https://github.com/prometheus/client_model/blob/d56cd794bca9543da8cc93e95432cd64d5f99635/io/prometheus/client/metrics.proto#L115-L122) float and integer numbers are explicitly transferred in different fields.
+
+### Exemplars
+
+* Reuse the same logic as with exponential histograms. Roughly: parse the exemplars from the classic histogram, discard exemplars without timestamp and sort by timestamp before forwarding.
 
 ### Writing/storage
 
--- a/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
+++ b/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
@@ -69,7 +69,7 @@ The counters for the custom buckets should be stored as integer values if possib
 
 ### Option 4 no suffix in series name
 
-> Options 1,2,3 are in the original Google [doc](https://docs.google.com/document/d/1aWNKUagsqUNlZWNjowOItx9JPueyuZ0ZWPyAhq6M8I8/edit#heading=h.kdvznfdr2b0o).
+> Options 1,2,3 are in the original Google [doc](https://docs.google.com/document/d/1aWNKUagsqUNlZWNjowOItx9JPueyuZ0ZWPyAhq6M8I8/edit#heading=h.es6bovk9enxw).
 
 In this option only instrumentation backwards compatibility is guaranteed(*). The user needs to switch to native histogram style queries, which makes this solution forward compatible with exponential histograms. Documented for example [here](https://grafana.com/docs/mimir/latest/visualize/native-histograms/#prometheus-query-language). In a later stage we could add an emulation layer on top of this to get backwards compatible queries.
 
@@ -135,7 +135,7 @@ The `histogram.Histogram` and `histogram.FloatHistogram` types to get an additio
 
 Chunk representation
 
-TBD, leaving to [beorn](beorn@grafana.com), but we can assume that the current integer histogram and float histogram chunk is just expanded with custom bucket definitions encoded after headers if the schema is set.
+We propose that the current integer histogram and float histogram chunk is expanded with custom bucket field encoding (and type is encoded in schema). Leaving exact format to PR author.
 
 Chunk iterator
 
--- a/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
+++ b/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
@@ -69,7 +69,7 @@ The counters for the custom buckets should be stored as integer values if possib
 
 ### Option 4 no suffix in series name
 
-> Options 1,2,3 are in the original Google [doc](https://docs.google.com/document/d/1aWNKUagsqUNlZWNjowOItx9JPueyuZ0ZWPyAhq6M8I8/edit#heading=h.es6bovk9enxw).
+> Options 1,2,3 are at the back of the document.
 
 In this option only instrumentation backwards compatibility is guaranteed(*). The user needs to switch to native histogram style queries, which makes this solution forward compatible with exponential histograms. Documented for example [here](https://grafana.com/docs/mimir/latest/visualize/native-histograms/#prometheus-query-language). In a later stage we could add an emulation layer on top of this to get backwards compatible queries.
 
@@ -168,7 +168,7 @@ No change to interface. The raw chunk iterator can load the custom bucket defini
 
 ## Open questions
 
-* How to enable the feature, what should be the name of the flags / configuration. Relation to native histograms feature.
+None.
 
 ## Answered questions
 
--- a/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
+++ b/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
@@ -13,6 +13,7 @@
   * [remote write 2.0 - explore histogram atomicity](https://github.com/prometheus/prometheus/issues/13410)
 
 * **Other docs or links:**
+  * [Epic](https://github.com/prometheus/prometheus/issues/13485) that contains the sub-tasks.
   * [Previous version of this proposal in Google doc](https://docs.google.com/document/d/1aWNKUagsqUNlZWNjowOItx9JPueyuZ0ZWPyAhq6M8I8/edit#heading=h.kdvznfdr2b0o)
   * [RW Self-Contained Histograms](https://docs.google.com/document/d/1mpcSWH1B82q-BtJza-eJ8xMLlKt6EJ9oFGH325vtY1Q/edit?pli=1#heading=h.ueg7q07wymku)
 
@@ -78,12 +79,12 @@ In this option only instrumentation backwards compatibility is guaranteed(*). Th
 ### Scrape configuration
 
 1. The feature is disabled if the feature flag [`native-histograms`](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms) is disabled.
-2. If native histograms feature is enabled, custom histograms can be enabled in `scrape_config` by setting the configration option `scrape_custom_histograms` to `true`.
+2. If native histograms feature is enabled, custom histograms can be enabled in `scrape_config` by setting the configration option `convert_classic_histograms` to `true`.
 
 Scenarios provided that the native histograms feature is enabled:
 * If the scraped metric has custom bucket definitions and `scrape_classic_histograms` is enabled, the original classic histogram series with suffixes shall be generated. This is independent of the setting for custom histograms.
-* If `scrape_custom_histograms` is disabled, no effect.
-* If `scrape_custom_histograms` is enabled:
+* If `convert_classic_histograms` is disabled, no effect.
+* If `convert_classic_histograms` is enabled:
   * If the scrape histogram has exponential buckets, no effect. Exponential buckets have higher priority.
   * Otherwise the new custom histograms are scraped in a series without suffix.
 
@@ -293,10 +294,3 @@ Option 4 (stage1) provides backward compatible instrumentation for the legacy sy
 * Migration path to exponential native histograms:
 * If feature is in use: 1. Enable native histograms.
 * If feature is not in use: skip custom buckets feature. 1. Update queries to query both classic histograms and native histograms. 2. Start scraping exponential histograms along classic histograms. 3. Stop scraping classic histograms.
-
-## Action Plan
-
-The tasks to do in order to migrate to the new idea.
-
-* [ ] Task one <GH issue>
-* [ ] Task two <GH issue> ...
--- a/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
+++ b/proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
@@ -78,7 +78,8 @@ In this option only instrumentation backwards compatibility is guaranteed(*). Th
 ### Scrape
 
 * Create the new representation in a series named `<metric>` without the `le` label. The resulting series name would be the same as if the user enabled exponential histograms. For example `http_request_latency_seconds_bucket`, `http_request_latency_seconds_count` and `http_request_latency_seconds_sum` will simply be stored as a one metric called `http_request_latency_seconds` from now on.
-* Resulting samples at scrape time reuse the existing data structures for native histograms, but use a special schema number to indicate custom buckets. A sample would have either custom bucket definitions or exponential schema and buckets, but not both. It would be the exponential buckets that would be preferred by scrape if both are present and custom buckets would be dropped.
+* Resulting samples at scrape time reuse the existing data structures for native histograms, but use a special schema number to indicate custom buckets. A sample would have either custom bucket definitions or exponential schema and buckets, but not both.
+* If the histogram has exponential buckets during scrape then only the exponential buckets are kept and the custom buckets would be dropped.
 * If all bucket counters and the overall counter of the histogram is determined to be a whole number, use integer histogram, otherwise use float histogram.
   * In the Prometheus text exposition format values are represented as floats. The dot (`.`) is not mandatory.
   * In the OpenMetrics text exposition [format](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#numbers) floats and integers are explicitly distinguished by the dot (`.`) sign being present or not in the value.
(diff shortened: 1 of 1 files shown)


bwplotka:
Nice! Will look in ~1h, thanks! 💪🏽

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,177 @@
...
+
+This proposal aims to minimize the changes needed to achieve the goals, but also let us optimize later.
+
+Enhance the internal representation of histograms (both float and [integer](https://github.com/prometheus/prometheus/blob/main/model/histogram/histogram.go)) with a nil-able slice of custom bucket definitions. No need to change spans/deltas/values slices.
+
+The counters for the custom buckets should be stored as integer values if possible. To be compatible with existing precision of the classic histogram representation within a to be defined 𝜎. The GO statement `x == math.Trunc(x)` has an error of around `1e-16` - experimentally.
+
+### Option 4 no suffix in series name
+
+> Options 1,2,3 are in the original Google [doc](https://docs.google.com/document/d/1aWNKUagsqUNlZWNjowOItx9JPueyuZ0ZWPyAhq6M8I8/edit#heading=h.kdvznfdr2b0o).
```
bwplotka, line 72:
```suggestion
> Options 1,2,3 are in the original Google [doc](https://docs.google.com/document/d/1aWNKUagsqUNlZWNjowOItx9JPueyuZ0ZWPyAhq6M8I8/edit#heading=h.es6bovk9enxw).
```

bwplotka (approved):
Nice! LGTM, some nits only.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,177 @@
...
+* Do we add a metric relabel config to govern the feature on a metric level?
+  *Answer:* NO. Already possible to achieve by enabling `scrape_classic_histograms` and selectively dropping metrics at scrape.
+* Would we want to implement the conversion from classic histogram to custom histogram on the remote write receiver side?
+  *Answer:* NO. It would require stateful receiver and probably transaction support on sender side, both of which we’d want to avoid.
+* Would we want to make the remote write endpoint configurable in storing the custom histogram? Potentially it could resolve the custom histogram into a classic histogram and store the series as separate again to only solve for the atomicity , but not performance.
+  *Answer:* YES, but not in the first iteration, not part of MVP.
+
+## Alternatives
+
+> Options 1,2,3 and a comparision is in the original Google [doc](https://docs.google.com/document/d/1aWNKUagsqUNlZWNjowOItx9JPueyuZ0ZWPyAhq6M8I8/edit#heading=h.kdvznfdr2b0o).
```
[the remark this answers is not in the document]
krajorama: You're right, I tried to make a shortcut, will copy

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Action Plan
+
+The tasks to do in order to migrate to the new idea.
+
+* [ ] Task one <GH issue>
+* [ ] Task two <GH issue> ...
```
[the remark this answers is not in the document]
krajorama: ack

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,177 @@
...
+* The function histogram_count would now potentially select custom bucket and exponential histogram samples at the same time. The interpretation of count is the same for custom and exponential histograms so this is not a problem.
+* The function histogram_sum would now potentially select custom bucket and exponential histogram samples at the same time. The interpretation of count is the same for custom and exponential histograms so this is not a problem.
+
+### Migration path for users
+
+Related CNCF slack [thread](https://cloud-native.slack.com/archives/C02KR205UMU/p1706087513992019?thread_ts=1706019629.852979&cid=C02KR205UMU).
+
+Use case: neither custom histograms nor exponential histograms in use (i.e. both features are off).
+
+* Enable feature to store custom bucket native histograms, scrape them in parallel to classic (via `scrape_classic_histograms` option), and do nothing else for a while (i.e. production usage still uses classic, change no dashboards and alerts.
```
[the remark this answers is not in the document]
krajorama: YES those are the options we a proposing currently, although we have not talked about the granularity of control? Is this a feature flag? A scrape level flag? A metric level flag?

> > Once you are done, switch off scraping classic histograms (or drop via metric relabel as needed).
> 
> Ah... hm. This is bit prone to mistakes though. Ideally option to convert to custom have configuration to delete classics automatically 🤔 We can decide on this later on (implementation detail a bit)

I don't understand this sentence, can you give an example?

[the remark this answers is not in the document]
bwplotka: I mean we want in future to have people drop classic histograms always, right? So dropping classic metrics will be a default. If for this you need manual relabelling for this (prone to error as -- making sure to not accidentally drop non classic histogram metrics and include all with _sum and _count etc), for many it will be simply to difficult to achieve. I think we need dedicated mode for that "removal". WDYT?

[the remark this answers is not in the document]
krajorama: Yes, the proposal is straightforward and drops all classic metrics by default. It doesn't depend on naming/relabelling, it's looking at the type, see [link](https://github.com/prometheus/prometheus/blob/fb552d290d8935c196a724a49d763080f3d7ff3a/model/textparse/protobufparse.go#L179). 

Conversely you can enable scraping all classic metrics with a single option.

However in reality these migration take a long time, so if you enabled custom histograms and scrape classic series at the same time, you might want to start dropping some classic series (esp. if they are high cardinality) and in that case you can add a relabel rule to drop some `<metric>_(bucket|count|sum)`. We don't actually have to do anything other then mention in documentation , since this is an existing feature.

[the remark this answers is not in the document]
bwplotka: Thanks! makes sense

bwplotka (approved):
Good to go from my side

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Pitfalls of the current solution
+
+* Classic histograms are split into time series and remote writes are not atomic above series level.
+* TCO (for some users this matters more than exponential buckets of full on native histograms, which require a migration effort to use)
+
+## Goals
+
+* No change to instrumentation.
+* No change to the query side. *Might not be achieved in the first iteration/ever. The ingestion and storage part can be fully implemented without any changes to the query part. A compatibility layer for querying can be introduced later as needed.*
```
SuperQ, line 38: I almost feel like this should be a non-goal. If we're going to convert classic to native, I feel like users that opt-in to such a feature would want all the normal behaviors of native histograms.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,302 @@
...
+
+* New instrumentation for defining the custom buckets.
+* Interoperability with (exponential bucket) native histograms.
+
+## How
+
+### Naming convention
+
+For the sake of brevity will use the following wording:
+* Classic histogram = classic histogram implying separate series for buckets, count and sum.
```
SuperQ, line 57: Minor wording improvment

```suggestion
* Classic histogram = Original Prometheus histogram implemented by separate series of buckets metrics labeled by `le`, with separate `_count` and `_sum` metrics.
```

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,302 @@
...
+> Options 1,2,3 are at the back of the document.
+
+In this option only instrumentation backwards compatibility is guaranteed(*). The user needs to switch to native histogram style queries, which makes this solution forward compatible with exponential histograms. Documented for example [here](https://grafana.com/docs/mimir/latest/visualize/native-histograms/#prometheus-query-language). In a later stage we could add an emulation layer on top of this to get backwards compatible queries.
+
+*: there is still the option to keep scraping classic histograms series beside custom histograms to be able to use the old queries, in fact it’s a recommended migration tactic, see later.
+
+### Scrape configuration
+
+1. The feature is disabled if the feature flag [`native-histograms`](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms) is disabled.
+2. If native histograms feature is enabled, custom histograms can be enabled in `scrape_config` by setting the configration option `scrape_custom_histograms` to `true`.
```
SuperQ, line 81: The naming here doesn't really give any clue that classic histograms are converted. I would have a scrape config feature named `convert_classic_histograms`.

SuperQ (commented):
While I see the desire to provide scrape config level flexibility. I think from an operator perspective I would prefer this be a global, per-Prometheus, option. I would not want inconsistent query or data behavior on the same instance of Prometheus. Some jobs emitting classic histograms, some jobs emitting native histograms. It's hard enough already with the existing classic histograms.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,302 @@
...
+* Remote write protocol to support transferring custom bucket definitions as part of the current Histogram DTO.
+* TSDB should reuse the existing chunk types.
+  * If the sample has exponential schema and buckets, store the exponential schema and buckets as currently.
+  * Otherwise if the sample has custom bucket definition and buckets, use a special schema number and store the custom bucket definitions at the beginning of the chunk once and then the bucket values as before. In the rare case when the bucket layout changes, start a new chunk to only store the custom bucket definitions once per chunk.
+
+### Reading via PromQL
+
+* Direct series selection of `metric{}` would look up `metric{}` and return the exponential histograms or custom histograms as found in the chunks.
+
+  > Note: the [exposition format](https://prometheus.io/docs/prometheus/latest/querying/api/#native-histograms) of histograms in PromQL is already forward compatible with custom buckets as the data contains the boundaries and inclusion settings.
```
krajorama, line 117: OTOH the exposition format would contain a log of redundant information especially for range query (matrix) results and in cases where the custom buckets don't change very often.

Maybe a better approach would be to change the exposition format from:
```
"histograms": [ [ <unix_time>, <histogram> ], ... ]

<histogram> = {
  "count": "<count_of_observations>",
  "sum": "<sum_of_observations>",
  "buckets": [ [ <boundary_rule>, "<left_boundary>", "<right_boundary>", "<count_in_bucket>" ], ... ]
}
```
to
```
"histograms": {
  "bounds": {
     0: [ <boundary_rule>, "<left_boundary>", "<right_boundary>"],
     1: [ <boundary_rule>, "<left_boundary>", "<right_boundary>"]
     ...
  },
  "values": [ [<timestamp>, <histogram>], ... ]

<histogram> = {
    "count": "<count_of_observations>",
    "sum": "<sum_of_observations>",
    "buckets": [ ["<boundary_key>", "<count_in_bucket>" ], ... ]
  }
}
```

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,302 @@
...
+* **Contributors:**
+  * Bartłomiej (Bartek) Płotka[@bwplotka](https://github.com/bwplotka) [bwplotka@gmail.com](mailto:bwplotka@gmail.com)
+
+* **Implementation Status:** Not implemented
+
+* **Related Issues and PRs:**
+  * [histograms: custom bucket layouts](https://github.com/prometheus/prometheus/issues/11277)
+  * [remote write 2.0 - explore histogram atomicity](https://github.com/prometheus/prometheus/issues/13410)
+
+* **Other docs or links:**
```
krajorama, line 15: Add link to meta EPIC: https://github.com/prometheus/prometheus/issues/13485

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
krajorama: This would also solve an issue with visualization, because without the knowledge about the bucket layout, the UI cannot correctly guess what the possible buckets are. For exponential histograms we should also make sure that 0, 1 bounds are consecutive so that the UI can know the scale, or just explicitly have a key called "scale_factor" , "exponent", whatever.

krajorama: Another idea that's maybe more self contained:
```
<histograms> = [ [ <unix_time>, <histogram> ], ... ]

<histogram> = {
  "count": "<count_of_observations>",
  "sum": "<sum_of_observations>",
  "bounds": [ [ <boundary_rule>, "<left_boundary>", "<right_boundary>"], ... ]
  "counts": [ [<bucket_count>], ...]
}
```

Where "bounds" could be omitted if the "bound" of the sample are equal to the previous "bounds".

This would allow a change in schema/resolution inside one series. It could also still encode scale factor at the end of the "bounds" list, since "counts" would only look up indexes bellow the size of "bounds".

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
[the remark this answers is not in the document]
krajorama: adding link to EPIC, so I'm removing this

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,177 @@
...
+## Open questions
+
+* How to enable the feature, what should be the name of the flags / configuration. Relation to native histograms feature.
+
+## Answered questions
+
+* Would we ever want to store the old representation and the new one at the same time?
+  *Answer:* YES. Already should work via the `existing scrape_classic_histograms` option.
+* What to do in queries if custom bucket and native histogram meet?
+  *Answer:* same as today with float vs native histogram.
```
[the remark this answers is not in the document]
krajorama: I've provided more context, not sure if this can be resolved now?

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
bwplotka: I think those are solid ideas, but we probably can postpone it until we the effect of not ideally optimized query/query range API results here. This is also true for exponential buckets, no?

bwplotka (approved):

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
krajorama: Yes, it would be applicable to exponential buckets as well. Probably less effective, but hopefully not worse. I'd have to calculate how many times a schema change needs to happen before this becomes an overhead.

krajorama: From the histograms weekly:  let's postpone this conversation, not part of MVP. Also we're talking about JSON which is verbose anyway, maybe more of an idea for Arrow/ProtoBuf serializations.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+
+This proposal aims to minimize the changes needed to achieve the goals, but also let us optimize later.
+
+Enhance the internal representation of histograms (both float and [integer](https://github.com/prometheus/prometheus/blob/main/model/histogram/histogram.go)) with a nil-able slice of custom bucket definitions. No need to change spans/deltas/values slices.
+
+The counters for the custom buckets should be stored as integer values if possible. To be compatible with existing precision of the classic histogram representation within a to be defined 𝜎. The GO statement `x == math.Trunc(x)` has an error of around `1e-16` - experimentally.
+
+### Option 4 no suffix in series name
+
+> Options 1,2,3 are at the back of the document.
```
SuperQ, line 73: This wording is slightly confusing.

```suggestion
### Naming: no suffix in series name

> Additional naming alternatives are at the end of the document.
```

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
beorn7: Let's introduce a 3rd section (next to _Goals_ and _Non-goals-): _Maybe-goals_. :)

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Audience
+
+* Operator, administrator of Prometheus infrastructure that deals with scraping and storing data.
+* Users of the query and remote read interface.
+
+## Non-Goals
+
+* New instrumentation for defining the custom buckets.
+* Interoperability with (exponential bucket) native histograms.
```
beorn7, line 51: That could be a "maybe-goal", too.

I.e. in the (unlikely) case that a bucketing schema from a classic histogram is convertible into something that can be merged with a exponential native histogram, the PromQL engine detect that and do the merge.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+* Exponential histogram = a native histogram with exponential buckets.
+* Custom histogram = a native histogram with custom (i.e. user defined) buckets.
+
+### Overview
+
+This proposal aims to minimize the changes needed to achieve the goals, but also let us optimize later.
+
+Enhance the internal representation of histograms (both float and [integer](https://github.com/prometheus/prometheus/blob/main/model/histogram/histogram.go)) with a nil-able slice of custom bucket definitions. No need to change spans/deltas/values slices.
+
+The counters for the custom buckets should be stored as integer values if possible. To be compatible with existing precision of the classic histogram representation within a to be defined 𝜎. The GO statement `x == math.Trunc(x)` has an error of around `1e-16` - experimentally.
```
beorn7, line 69: Hmm, is that really required?

For one, exposition formats make it quite clear if a histogram is integer or float. It is explicit in protobuf. OpenMetrics text format currently doesn't even support float histograms. And assuming it does so in the future, the floats are explicitly marked by the presence of dots.

Classic Prometheus text format isn't explicit about it, but we could still say we ingest as float histogram, if at least one bucket or the count has a dot _and_ has non-zero digits after the dot.

There is the edge case of explicitly marked float histograms that are effectively integer histograms because all contained floats have only zeros after the floating point. Currently, we store those histograms as float histograms if they are native histograms, and that's probably fine as it is rare. But we do have the plan to optimize that case and opportunistically convert them to an integer histogram. We would only do so if every involved float is effectively an integer.

The way IEEE 754 floats work, this is always clear. We don't need any rounding. If you convert an integer to a float, then `math.Trunc(x) == x`, even if you are beyond the range where float64 represents each integer precisely (±2^53).

This is also how we did it in Prometheus v1 (which opportunistically stored the simple float64 samples as integers if they were effectively (and precisely) integers).

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+> Additional naming alternatives are at the end of the document.
+
+In this option only instrumentation backwards compatibility is guaranteed(*). The user needs to switch to native histogram style queries, which makes this solution forward compatible with exponential histograms. Documented for example [here](https://grafana.com/docs/mimir/latest/visualize/native-histograms/#prometheus-query-language). In a later stage we could add an emulation layer on top of this to get backwards compatible queries.
+
+*: there is still the option to keep scraping classic histograms series beside custom histograms to be able to use the old queries, in fact it’s a recommended migration tactic, see later.
+
+### Scrape configuration
+
+1. The feature is disabled if the feature flag [`native-histograms`](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms) is disabled.
+2. If native histograms feature is enabled, custom histograms can be enabled in `scrape_config` by setting the configration option `convert_classic_histograms` to `true`.
```
beorn7, line 82: To address @SuperQ 's comment: I think there should definitely be an option per-scrape-config, but we could also provide a global option for those that want the feature always.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+
+Scenarios provided that the native histograms feature is enabled:
+* If the scraped metric has custom bucket definitions and `scrape_classic_histograms` is enabled, the original classic histogram series with suffixes shall be generated. This is independent of the setting for custom histograms.
+* If `convert_classic_histograms` is disabled, no effect.
+* If `convert_classic_histograms` is enabled:
+  * If the scrape histogram has exponential buckets, no effect. Exponential buckets have higher priority.
+  * Otherwise the new custom histograms are scraped in a series without suffix.
+
+> Note: enabling native histograms comes with all notes and warnings described in the documentation, in particular the impact on the `le` label for the classic histograms for certain clients. See "Note about the format of le and quantile label values" in [Native Histograms](https://prometheus.io/docs/prometheus/latest/feature_flags/#native-histograms).
```
beorn7, line 91: But interestingly, when scraping classic histograms into native histograms with custom buckets, you got normalization of the `le` label for free.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+* The function histogram_count would now potentially select custom bucket and exponential histogram samples at the same time. The interpretation of count is the same for custom and exponential histograms so this is not a problem.
+* The function histogram_sum would now potentially select custom bucket and exponential histogram samples at the same time. The interpretation of count is the same for custom and exponential histograms so this is not a problem.
+
+### Migration path for users
+
+Related CNCF slack [thread](https://cloud-native.slack.com/archives/C02KR205UMU/p1706087513992019?thread_ts=1706019629.852979&cid=C02KR205UMU).
+
+Use case: neither custom histograms nor exponential histograms in use (i.e. both features are off).
+
+* Enable feature to store custom bucket native histograms, scrape them in parallel to classic (via `scrape_classic_histograms` option), and do nothing else for a while (i.e. production usage still uses classic, change no dashboards and alerts.
```
beorn7, line 132:
```suggestion
* Enable feature to store custom bucket native histograms, scrape them in parallel to classic (via `scrape_classic_histograms` option), and do nothing else for a while (i.e. production usage still uses classic, change no dashboards and alerts).
```

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+
+The `message.Histogram` type to be expanded with nullable repeated custom_buckets field that list the custom bucket definitions (except `+Inf`, which is implicit). There should be a comment which specifies which schema number means that we need to even look at this field. It should be a validation error to find this field null if the custom bucket schema number is used.
+
+Internal representation
+
+The `histogram.Histogram` and `histogram.FloatHistogram` types to get an additional field that has the custom bucket definitions or reference id for interned custom bucket definition. Doesn’t matter as it should be accessible via getter interface. (We’d have to see how the PromQL engine uses the bucket definitions to nail down what exact interface to put here.)
+
+Chunk representation
+
+We propose that the current integer histogram and float histogram chunk is expanded with custom bucket field encoding (and type is encoded in schema). Leaving exact format to PR author.
```
beorn7, line 164: The chunk already contains a `schema` field in the beginning, doesn't it? That field will dictate the data layout following it (i.e. is there a list of bucket definitions or not).

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+* Configuration option to be documented.
+* Impact on queries and PromQL functions to be specified.
+
+### Data structures
+
+New constant to define the schema number to mean custom buckets (e.g. 127 ?).
+
+Remote write protocol
+
+The `message.Histogram` type to be expanded with nullable repeated custom_buckets field that list the custom bucket definitions (except `+Inf`, which is implicit). There should be a comment which specifies which schema number means that we need to even look at this field. It should be a validation error to find this field null if the custom bucket schema number is used.
```
beorn7, line 156: It should be noted that we don't have to specify that the last boundary is `+Inf` (as that's always the case) but we _should_ store a count for the final `+Inf` bucket. In current classic histograms, that value is always equal to the count value, but in native histogram, we specify that observations of `NaN` increase the count, but not any bucket.

Once the custom buckets are implemented, we can in principle use them as pure native histograms with custom buckets on the instrumentation side, and then we would also get this behavior.

(In different news, I think it is a design flaw of classic histograms that they effectively count NaN observations in the +Inf bucket. On the other hand, it doesn't really matter in practices, but now that it is cleanly implemented in native histograms, we should keep that behavior throughout.)

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Open questions
+
+None.
+
+## Answered questions
+
+* Would we ever want to store the old representation and the new one at the same time?
+  *Answer:* YES. Already should work via the `existing scrape_classic_histograms` option.
+* What to do in queries if custom histogram and exponential histogram meet or customer histogram and float sample?
```
beorn7, line 178:
```suggestion
* What to do in queries if custom histogram and exponential histogram meet or custom histogram and float sample?
```

And I would also add "custom histograms of different layout".

SuperQ:
Nice

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+  *Answer:* same as today with float vs native histogram, that is calculate the result if it makes mathematical sense. For example multiplying a custom histogram with the number 2.0 makes sense. In case of histograms they need to rescaled to match their schema.
```
beorn7, line 179: "rescaled" is confusing (as we also use it if we multiply a histogram with a float). Maybe "change resolution" or something. It should also be added that this might not work precisely, and that in that case we rather give a warning and don't do the calculation instead of doing something wonky.

Examples (not necessarily to add here, just to create understanding for the reviewers):

* Custom histogram with boundaries 1, 2, 4, 8 and no sample in the +Inf bucket → can be merged with any exponential histogram.
* Anything in the +Inf bucket: Doesn't work with exponential histogram.
* Custom histogram with boundaries 1, 5, 10, 20, +Inf added to another custom histogram with 1, 10, 20, +Inf: We can do the math by first merging the 5-bucket and the 10-bucket together in the first histogram and then add both, resulting in a 1, 10, 20 histogram.

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
```diff
@@ -0,0 +1,296 @@
...
+* Should we use a bigger chunk size for such custom histograms? To offset that we’d want to store the bucket layout in the chunk header. ~4K?
+  *Answer:* NO. Classic histograms typically have less buckets than exponential native histograms which should offset any additional information encoded in the chunk.
```
beorn7, line 181:
```suggestion
  *Answer:* NO. Classic histograms typically have fewer buckets than exponential native histograms, which should offset any additional information encoded in the chunk.
```

beorn7 (changes requested):

beorn7:
This was merged during my review.

I recommend to still read my thoughtful comments and apply them as needed. :pray:

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
krajorama: i don't understand this, it's not free, because you might have to implement the relabel trick we documented

beorn7: What trick are you referring to?

proposals/2024-01-26_classic-histograms-stored-as-native-histograms.md
krajorama: yes that's what I meant, clarifying

PR #32, opened 2024-01-29
Proposal for Native Histogram Text Format

csmarchbanks:
Create an initial prosal for supporting native histograms in the text format. It mainly links to the design doc for ease of comments while we decide on a path forward. After a proposal from the design doc has some support the rest of the markdown proposal will be filled out.

Leaving the PR as a draft until the markdown is ready to be published, but comments here or in the Google doc are welcome.

### DIFF
--- a/proposals/2024-01-29_native_histograms_text_format.md
+++ b/proposals/2024-01-29_native_histograms_text_format.md
@@ -31,7 +31,7 @@ Goals and use cases for the solution as proposed in [How](#how):
 * Support native histograms in the text format
 * (Secondary) Encode/decode efficiency
 * (Secondary) Ease of implementation for client libraries
-* (Secondary) Human readibility of the format
+* (Secondary) Human readability of the format
 
 Note that the goals of efficiency and human readability are commonly at odds with each other.
 
@@ -46,7 +46,7 @@ Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
 
 ## How
 
-Extend the OpenMetrics text format to allow structured values instead of only float values. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto). Starting with examples and then breaking up the format:
+Extend the OpenMetrics text format to allow structured values instead of only float values for specific series of a histogram type. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto). Starting with examples and then breaking up the format:
 ```
 # TYPE nativehistogram histogram
 nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
@@ -61,18 +61,21 @@ hist_with_classic_buckets_bucket{le="+Inf"} 24
 hist_with_classic_buckets_count 24
 hist_with_classic_buckets_sum 100
 ```
-The metric will have no "magic" suffixes, then the value for each series is a custom struct format with the following fields:
+
+Native histograms will share the "histogram" type with classic histograms. Classic and native histograms can be differentiated by looking at the "magic" suffixes for classic histogram series (`_bucket`, `_count`, `_sum`), and no suffix for native histogram series. This allows producers to expose native histograms and classic histograms together if desired, such as desiring custom bucket boundaries. An optional `_created` series can be created if desired just like a classic histogram as well.
+
+The value for each series of a native histogram is a custom struct format with the following fields inside curly braces:
 * `sum: float64` - The sum of all observations for this histogram. Could be negative in cases with negative observations.
 * `count: uint64` - The number of samples that were observed for this histogram.
 * `schema: int32` - The schema used for this histogram, currently supported values are -4 -> 8.
 * `zero_threshold: float64` - The width of the zero bucket.
 * `zero_count: uint64` - The number of observations inside the zero bucket.
 * `negative_spans: []BucketSpan` - The buckets corresponding to negative observations, optional.
-* `negative_deltas: []int64` - The delta of counts compared to the previous bucket. 
+* `negative_deltas: []int64` - The delta of counts compared to the previous bucket, optional.
 * `positive_spans: []BucketSpan` - The buckets corresponding to negative observations, optional.
-* `positive_deltas: []int64` - The delta of counts compared to the previous bucket. 
+* `positive_deltas: []int64` - The delta of counts compared to the previous bucket, optional.
 
-A bucket span is the combination of an `int32` offset and a `uint32` length. It is encoded as `<offset>:<length>`. Lists/arrays are encoded within square brackets with elements separated by commas. Compared to JSON this avoids consistently repeating keys and curly braces.
+A bucket span is the combination of an `int32` offset and a `uint32` length. It is encoded as `<offset>:<length>`. Lists/arrays are encoded within square brackets with elements separated by commas. Compared to JSON this avoids consistently repeating keys and curly braces. White space is not allowed inside of the structure to make a value as easy as possible to parse.
 
 Positive infinity, negative infinity, and non number values will be represented as case insensitive versions of `+Inf`, `-Inf`, and `NaN` respectively in any field. This is the same behavior for values in OpenMetrics today.
 
--- a/proposals/2024-01-29_native_histograms_text_format.md
+++ b/proposals/2024-01-29_native_histograms_text_format.md
@@ -16,9 +16,9 @@
 
 ## Why
 
-Today it is only possible to export native histograms using the Protocol Buffers (protobuf) scrape format. Many users prefer the text format, and some client libraries, such as the Python client, want to avoid adding a dependency on protobuf. 
+Today it is only possible to export native histograms using the classic (not OpenMetrics) Protocol Buffers (protobuf) scrape format. Many users prefer the text format, and some client libraries, such as the Python client, want to avoid adding a dependency on protobuf. Prometheus content negotiation prefers OpenMetrics to the classic Prometheus text based format, therefore to support native histograms in Prometheus the OpenMetrics text format will also need to support native histograms.
 
-During a [dev summit in 2022](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#bookmark=id.c3e7ur6rn5d2) there was consensus we would continue to support the text format. Including native histograms as part of the text format shows commitment to that consensus.
+There is already an open pull request (see Related Issues and PRs above) to add support for native histograms to OpenMetrics, and during a [dev summit in 2022](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#bookmark=id.c3e7ur6rn5d2) there was consensus we would continue to support the text format for new features as well. Including native histograms as part of the text format shows commitment to that consensus.
 
 ### Pitfalls of the current solution
 
@@ -42,13 +42,24 @@ Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
 ## Non-Goals
 
 * Requiring backwards compatability (OpenMetrics 2.0 would be ok), and especially forwards compatability (not required in the OpenMetrics spec).
+* Support for multiple exemplars (this will be done in a future proposal).
 
 ## How
 
-Extend the OpenMetrics text format to allow structured values instead of only float values. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/OpenObservability/OpenMetrics/pull/256). Starting with an example and then breaking up the format:
+Extend the OpenMetrics text format to allow structured values instead of only float values. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto). Starting with examples and then breaking up the format:
 ```
 # TYPE nativehistogram histogram
-nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_span:[0:2,1:2],negative_span:[0:2,1:2],positive_delta:[2,1,-2,3],negative_delta:[2,1,-2,3]}
+nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+
+# TYPE hist_with_labels histogram
+hist_with_labels{foo="bar",baz="qux"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+
+# TYPE hist_with_classic_buckets histogram
+hist_with_classic_buckets {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+hist_with_classic_buckets_bucket{le="0.001"} 4
+hist_with_classic_buckets_bucket{le="+Inf"} 24
+hist_with_classic_buckets_count 24
+hist_with_classic_buckets_sum 100
 ```
 The metric will have no "magic" suffixes, then the value for each series is a custom struct format with the following fields:
 * `sum: float64` - The sum of all observations for this histogram. Could be negative in cases with negative observations.
@@ -56,20 +67,29 @@ The metric will have no "magic" suffixes, then the value for each series is a cu
 * `schema: int32` - The schema used for this histogram, currently supported values are -4 -> 8.
 * `zero_threshold: float64` - The width of the zero bucket.
 * `zero_count: uint64` - The number of observations inside the zero bucket.
-* `negative_span: []BucketSpan` - The buckets corresponding to negative observations, optional.
-* `negative_delta: []int64` - The delta of counts compared to the previous bucket. 
-* `positive_span: []BucketSpan` - The buckets corresponding to negative observations, optional.
-* `positive_delta: []int64` - The delta of counts compared to the previous bucket. 
+* `negative_spans: []BucketSpan` - The buckets corresponding to negative observations, optional.
+* `negative_deltas: []int64` - The delta of counts compared to the previous bucket. 
+* `positive_spans: []BucketSpan` - The buckets corresponding to negative observations, optional.
+* `positive_deltas: []int64` - The delta of counts compared to the previous bucket. 
 
 A bucket span is the combination of an `int32` offset and a `uint32` length. It is encoded as `<offset>:<length>`. Lists/arrays are encoded within square brackets with elements separated by commas. Compared to JSON this avoids consistently repeating keys and curly braces.
 
 Positive infinity, negative infinity, and non number values will be represented as case insensitive versions of `+Inf`, `-Inf`, and `NaN` respectively in any field. This is the same behavior for values in OpenMetrics today.
 
-Note that in this initial implementation float histograms are not supported.
+Note that in this initial implementation float histograms are not supported. Float histograms are rarely used in exposition, and OpenMetrics does not support classic float histograms either. Support could be added in the future by adding fields for `count_float`, `zero_count_float`, `negative_counts`, and `positive_counts`.
 
-### Backwards compatability and semantic versioning
+### Backwards compatibility and semantic versioning
 
 After discussions with a few people it is believed that these changes can be made in a 1.x release of OpenMetrics. OpenMetrics 1.x parsers that support native histograms will still be able to read OpenMetrics 1.0 responses, therefore this change is backwards compatible. However, this change is not forwards compatible, i.e. an OpenMetrics 1.0 parser will not be able to read an OpenMetrics >= 1.1 response. Any producers implementing native histograms MUST also implement content negotiation and fall back to OpenMetrics 1.0.0, and therefore not expose native histograms, if a supported version cannot be negotiated. Note that the behavior to fall back to 1.0.0 is already part of the [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#protocol-negotiation).
+
+Until a version of OpenMetrics is released that contains a stable version of native histograms consumers can determine if native histograms may be present by asking for a `nativehistogram` pre-release identifier. For example,
+```
+Accept: application/openmetrics-text;version=1.1.0-nativehistogram.*,application/openmetrics-text;version=1.0.0,text/plain;version=0.0.4
+```
+would mean the consumer can accept any nativehistogram enabled pre-release version of OpenMetrics 1.1.0, the base 1.0.0 version of OpenMetrics, or the 0.0.4 version of the classic Prometheus text format. Producers must include the proper content type for their version, such as the first nativehistogram pre-release:
+```
+Content-Type: application/openmetrics-text;version=1.1.0-nativehistogram.0
+```
 
 ## Alternatives
 
--- a/proposals/2024-01-29_native_histograms_text_format.md
+++ b/proposals/2024-01-29_native_histograms_text_format.md
@@ -42,6 +42,8 @@ Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
 ## Non-Goals
 
 * Requiring backwards compatability (OpenMetrics 2.0 would be ok), and especially forwards compatability (not required in the OpenMetrics spec).
+* Support for custom buckets as part of the native histogram structure. If needed by a consumer, custom buckets will need to be parsed from classic buckets.
+* Changing created timestamp semantics. Created timestamps will continue to be exposed as an additional series as with OpenMetrics 1.0.
 
 ## How
 
--- a/proposals/2024-01-29_native_histograms_text_format.md
+++ b/proposals/2024-01-29_native_histograms_text_format.md
@@ -1,0 +1,60 @@
+## Native Histograms Text Format
+
+* **Owners:**
+  * Chris Marchbanks [@csmarchbanks](https://github.com/csmarchbanks)
+
+* **Implementation Status:** Not implemented
+
+* **Related Issues and PRs:**
+  * `<GH Issues/PRs>`
+
+* **Other docs or links:**
+  * [Design Doc for choosing a proposal](https://docs.google.com/document/d/1qoHf24cKMpa1QHskIjgzyf3oFhIPvacyJj8Tbe6fIrY/edit#heading=h.5sybau7waq2q)
+
+> TL;DR: This design doc is proposing a format for exposing native histograms in the OpenMetrics text format.
+
+## Why
+
+Today it is only possible to export native histograms using the protobuf scrape format. Many users prefer the text format, and some client libraries, such as the Python client, want to avoid adding a dependency on protobuf. 
+
+During a [dev summit in 2022](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#bookmark=id.c3e7ur6rn5d2) there was consensus we would continue to support the text format. Including native histograms as part of the text format shows commitment to that consensus.
+
+See the linked design doc in Google Docs for additional background information.
+
+### Pitfalls of the current solution
+
+Prometheus client libraries such as Python do not want to require a dependency on protobuf in order to expose native histograms, and in some languages protobuf is painful to use. Gating native histograms only to clients/users willing to use protobuf hurts adoption of native histograms, therefore, we would like a way to represent a histogram in the text based format.
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Support native histograms in the text format
+* (Secondary) Encode/decode efficiency
+* (Secondary) Ease of implementation for client libraries
+* (Secondary) Human readibility of the format
+
+Note that the goals of efficiency and human readability are commonly at odds with each other.
+
+### Audience
+
+Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
+
+## Non-Goals
+
+* Requiring backwards compatability (OpenMetrics 2.0 would be ok)
+
+## How
+
+To be filled out after one of the proposals in the linked design doc has support.
+
+## Alternatives
+
+To be filled out after one of the proposals in the linked design doc has support.
+
+## Action Plan
+
+The tasks to do in order to migrate to the new idea.
+
+* [ ] Task one <GH issue>
+* [ ] Task two <GH issue> ...
--- a/proposals/2024-01-29_native_histograms_text_format.md
+++ b/proposals/2024-01-29_native_histograms_text_format.md
@@ -42,7 +42,6 @@ Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
 ## Non-Goals
 
 * Requiring backwards compatability (OpenMetrics 2.0 would be ok), and especially forwards compatability (not required in the OpenMetrics spec).
-* Support for multiple exemplars (this will be done in a future proposal).
 
 ## How
 
@@ -100,6 +99,13 @@ hist_with_classic_buckets_bucket{foo="baz",le="+Inf"} 24
 hist_with_classic_buckets_count{foo="baz"} 24
 hist_with_classic_buckets_sum{foo="baz"} 100
 hist_with_classic_buckets_created{foo="baz"} 1717536098
+```
+
+Finally, multiple exemplars will also be supported in the exposition format by providing a list of exemplars at the end of any line, separated by `#`. Note that having spaces around the hashes is required and matches the [ABNF specification in OpenMetrics](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#abnf). For example:
+```
+# TYPE exemplar_example histogram
+# HELP exemplar_example Is an example of a native histogram with exemplars.
+nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]} # {trace_id="KOO5S4vxi0o"} 0.67 # {trace_id="oHg5SJYRHA0"} 9.8 1520879607.789
 ```
 
 ### Backwards compatibility and semantic versioning
--- a/proposals/2024-01-29_native_histograms_text_format.md
+++ b/proposals/2024-01-29_native_histograms_text_format.md
@@ -49,12 +49,15 @@ Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
 Extend the OpenMetrics text format to allow structured values instead of only float values for specific series of a histogram type. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto). Starting with examples and then breaking up the format:
 ```
 # TYPE nativehistogram histogram
+# HELP nativehistogram Is a basic example of a native histogram.
 nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
 
 # TYPE hist_with_labels histogram
+# HELP hist_with_labels Is an example of a native histogram with labels.
 hist_with_labels{foo="bar",baz="qux"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
 
 # TYPE hist_with_classic_buckets histogram
+# HELP hist_with_classic_buckets Is an example of native and classic histograms together.
 hist_with_classic_buckets {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
 hist_with_classic_buckets_bucket{le="0.001"} 4
 hist_with_classic_buckets_bucket{le="+Inf"} 24
@@ -80,6 +83,24 @@ A bucket span is the combination of an `int32` offset and a `uint32` length. It 
 Positive infinity, negative infinity, and non number values will be represented as case insensitive versions of `+Inf`, `-Inf`, and `NaN` respectively in any field. This is the same behavior for values in OpenMetrics today.
 
 Note that in this initial implementation float histograms are not supported. Float histograms are rarely used in exposition, and OpenMetrics does not support classic float histograms either. Support could be added in the future by adding fields for `count_float`, `zero_count_float`, `negative_counts`, and `positive_counts`.
+
+If native histograms and a classic histogram are exposed simultaneously the native histogram must come first for any labelset. For example:
+```
+# TYPE hist_with_classic_buckets histogram
+# HELP hist_with_classic_buckets Is an example of combining native and classic histograms for two labelsets.
+hist_with_classic_buckets{foo="bar"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+hist_with_classic_buckets_bucket{foo="bar",le="0.001"} 4
+hist_with_classic_buckets_bucket{foo="bar",le="+Inf"} 24
+hist_with_classic_buckets_count{foo="bar"} 24
+hist_with_classic_buckets_sum{foo="bar"} 100
+hist_with_classic_buckets_created{foo="bar"} 1717536092
+hist_with_classic_buckets{foo="baz"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+hist_with_classic_buckets_bucket{foo="baz",le="0.001"} 4
+hist_with_classic_buckets_bucket{foo="baz",le="+Inf"} 24
+hist_with_classic_buckets_count{foo="baz"} 24
+hist_with_classic_buckets_sum{foo="baz"} 100
+hist_with_classic_buckets_created{foo="baz"} 1717536098
+```
 
 ### Backwards compatibility and semantic versioning
 
--- a/proposals/2024-01-29_native_histograms_text_format.md
+++ b/proposals/2024-01-29_native_histograms_text_format.md
@@ -41,7 +41,7 @@ Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
 
 ## Non-Goals
 
-* Requiring backwards compatability (OpenMetrics 2.0 would be ok)
+* Requiring backwards compatability (OpenMetrics 2.0 would be ok), and especially forwards compatability (not required in the OpenMetrics spec).
 
 ## How
 
@@ -66,6 +66,10 @@ A bucket span is the combination of an `int32` offset and a `uint32` length. It 
 Positive infinity, negative infinity, and non number values will be represented as case insensitive versions of `+Inf`, `-Inf`, and `NaN` respectively in any field. This is the same behavior for values in OpenMetrics today.
 
 Note that in this initial implementation float histograms are not supported.
+
+### Backwards compatability and semantic versioning
+
+After discussions with a few people it is believed that these changes can be made in a 1.x release of OpenMetrics. OpenMetrics 1.x parsers that support native histograms will still be able to read OpenMetrics 1.0 responses, therefore this change is backwards compatible. However, this change is not forwards compatible, i.e. an OpenMetrics 1.0 parser will not be able to read an OpenMetrics >= 1.1 response. Any producers implementing native histograms MUST also implement content negotiation and fall back to OpenMetrics 1.0.0, and therefore not expose native histograms, if a supported version cannot be negotiated. Note that the behavior to fall back to 1.0.0 is already part of the [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#protocol-negotiation).
 
 ## Alternatives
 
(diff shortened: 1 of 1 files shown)


proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,60 @@
...
+  * `<GH Issues/PRs>`
+
+* **Other docs or links:**
+  * [Design Doc for choosing a proposal](https://docs.google.com/document/d/1qoHf24cKMpa1QHskIjgzyf3oFhIPvacyJj8Tbe6fIrY/edit#heading=h.5sybau7waq2q)
+
+> TL;DR: This design doc is proposing a format for exposing native histograms in the OpenMetrics text format.
+
+## Why
+
+Today it is only possible to export native histograms using the protobuf scrape format. Many users prefer the text format, and some client libraries, such as the Python client, want to avoid adding a dependency on protobuf.
```
krajorama, line 18: nit: let's use the official name for Protocol Buffers once on top https://protobuf.dev/overview/

```suggestion
Today it is only possible to export native histograms using the Protocol buffers (protobuf) scrape format. Many users prefer the text format, and some client libraries, such as the Python client, want to avoid adding a dependency on protobuf. 
```

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,60 @@
...
+
+Today it is only possible to export native histograms using the protobuf scrape format. Many users prefer the text format, and some client libraries, such as the Python client, want to avoid adding a dependency on protobuf. 
+
+During a [dev summit in 2022](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#bookmark=id.c3e7ur6rn5d2) there was consensus we would continue to support the text format. Including native histograms as part of the text format shows commitment to that consensus.
+
+See the linked design doc in Google Docs for additional background information.
+
+### Pitfalls of the current solution
+
+Prometheus client libraries such as Python do not want to require a dependency on protobuf in order to expose native histograms, and in some languages protobuf is painful to use. Gating native histograms only to clients/users willing to use protobuf hurts adoption of native histograms, therefore, we would like a way to represent a histogram in the text based format.
```
krajorama, line 26: nit:
```suggestion
Prometheus client libraries such as Python do not want to require a dependency on protobuf in order to expose native histograms, and in some languages protobuf is painful to use. Gating native histograms only to clients/users willing to use protobuf hurts adoption of native histograms, therefore, we would like a way to represent a native histogram in the text based format.
```

csmarchbanks:
I have updated this proposal with the option from the design doc that gained consensus, and described the proposed format in specific detail. All reviews here are now welcome :)

ArthurSens (commented):
Hey 👋 

Do I understand correctly that the breaking change here is because you are using the same metric type as the classic histogram, but a different format for the exposed measurements?

If we're moving forward with OpenMetrics 2.0, we might want to align different workforces that are happening towards that, e.g. different format for created timestamps

csmarchbanks:
I think this could be done with a minor version bump of OpenMetrics, but if it requires going to 2.0 then yes we would coordinate with the other changes. Until then it would be behind feature flags and we would need a version like version=1.1.0-alpha.1 for content negotiation with clients.

ArthurSens (commented):
I wanted to approve the PR to give you the green light to experiment with things, but as mentioned in the dev summit I don't have the confidence to approve changes to OM.

My initial concerns are:
* Possible breaking change that requires OM 2.0 - You're using `# TYPE <metric name> histogram` followed by a line that has a different format from the current histogram format. Current scrapers will break when seeing this.

We could use `# TYPE <metric name> native_histogram` instead, and change it back to `histogram` in 2.0, if the scraper silently ignores unknown metric types, but I'm almost certain that's not the current behavior 😬 

* What feature flag are you planning to release this under? The existing `native-histogram` flag? A new one?

csmarchbanks:
Reasonable concerns! I added a section on backwards compatibility/versioning with the reason I believe this change to be backwards-compatible and releasable as part of OpenMetrics 1.1. Does that section help?

As far as feature flags go, I think reusing the same native-histogram flag would be the best option (plus content negotiation). I am ok with a new one as well though if others would like to see a new flag. That decision could probably happen at PR review time?

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,90 @@
...
+  * [OpenMetrics Protobuf format PR](https://github.com/OpenObservability/OpenMetrics/pull/256)
+
+* **Other docs or links:**
+  * [Design Doc for choosing a proposal](https://docs.google.com/document/d/1qoHf24cKMpa1QHskIjgzyf3oFhIPvacyJj8Tbe6fIrY/edit#heading=h.5sybau7waq2q)
+
+> TL;DR: This design doc is proposing a format for exposing native histograms in the OpenMetrics text format.
+
+## Why
+
+Today it is only possible to export native histograms using the Protocol Buffers (protobuf) scrape format. Many users prefer the text format, and some client libraries, such as the Python client, want to avoid adding a dependency on protobuf.
```
beorn7, line 19: To avoid confusion, can we clarify that the protobuf format used isn't even the OpenMetrics protobuf format, but the classic Prometheus protobuf format. (The proposed changes to support native histograms in the OpenMetrics protobuf format is linked above, so here we could explain that.)

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,90 @@
...
+
+Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
+
+## Non-Goals
+
+* Requiring backwards compatability (OpenMetrics 2.0 would be ok), and especially forwards compatability (not required in the OpenMetrics spec).
+
+## How
+
+Extend the OpenMetrics text format to allow structured values instead of only float values. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/OpenObservability/OpenMetrics/pull/256). Starting with an example and then breaking up the format:
```
beorn7, line 48: Maybe better link the protobuf spec that is actually used, i.e. https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,90 @@
...
+```
+# TYPE nativehistogram histogram
+nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_span:[0:2,1:2],negative_span:[0:2,1:2],positive_delta:[2,1,-2,3],negative_delta:[2,1,-2,3]}
```
beorn7, line 51: Protobuf infamously uses singular forms for arrays, but most readers of the text format won't be aware of that. Maybe it's better to use plural forms here? E.g. `positive_spans`, `negative_deltas`.

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,90 @@
...
+* `negative_span: []BucketSpan` - The buckets corresponding to negative observations, optional.
+* `negative_delta: []int64` - The delta of counts compared to the previous bucket. 
+* `positive_span: []BucketSpan` - The buckets corresponding to negative observations, optional.
+* `positive_delta: []int64` - The delta of counts compared to the previous bucket. 
+
+A bucket span is the combination of an `int32` offset and a `uint32` length. It is encoded as `<offset>:<length>`. Lists/arrays are encoded within square brackets with elements separated by commas. Compared to JSON this avoids consistently repeating keys and curly braces.
+
+Positive infinity, negative infinity, and non number values will be represented as case insensitive versions of `+Inf`, `-Inf`, and `NaN` respectively in any field. This is the same behavior for values in OpenMetrics today.
+
+Note that in this initial implementation float histograms are not supported.
```
beorn7, line 68: Maybe add some justification to this:
- Float histograms are rarely used in exposition.
- OpenMetrics currently doesn't support classic float histograms either.
- Float support would be easy to add (just need `count_float`, `zero_count_float`, `negative_counts`, `positive_counts` with the corresponding float types).

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Backwards compatability and semantic versioning
```
beorn7, line 70: compatability → compatibility

beorn7 (commented):
Looks great. Especially the clarifying part about the meaning of forward and backward compatibility in OM makes me happy.

I have just a few optional suggestions for additions.

Let's get a few more thumbs up on this and then approve.

beorn7:
> * Possible breaking change that requires OM 2.0 - You're using `# TYPE <metric name> histogram` followed by a line that has a different format from the current histogram format. Current scrapers will break when seeing this.
> 
> We could use `# TYPE <metric name> native_histogram` instead, and change it back to `histogram` in 2.0, if the scraper silently ignores unknown metric types, but I'm almost certain that's not the current behavior 😬

I'm pretty sure that OM is fairly strict with this. While you are free to implement a tolerant parser, it is not required. A fully OM 1.0 compliant parser may (and maybe even should) complain about any unknown metric types. A behavior like "new metric types maybe added, and old parsers should just ignore them" would need to be clearly specified in the spec. Cf. how protobuf is doing it.

The explanation of how forward and backward compatibility works as it is part of this proposal IMHO makes a lot of sense (and I wish it would have been explained that way in the OM spec). To paraphrase: A minor version bump is sufficient if a v1.1 scraper can read a v1.0 exposition without any precaution, i.e. it doesn't even have to know that it is reading v1.0 and not v1.1. It is OK and expected that a v1.0 scraper will be unable to parse a v1.1 exposition.

The linked explanation of content negotiation directly from the OM spec supports that: The scraper can ask for v1.1, but it might just get v1.0 back, which is fine. If on the other hand the scraper asks for v1.0, it must not get v1.1 back, but the producer must be capable of generating valid v1.0 and serve it as such (or, I guess, return an error).

> * What feature flag are you planning to release this under? The existing `native-histogram` flag? A new one?

Since this is just about the protocol, and there are many possible implementers (scrapers as well as producers), I won't discuss feature flags (which are specific to selected implementers, like in this case the Prometheus server).

However, it might be good to sketch out a plan how to version the implementation as long as there is no official OM v1.1. (Something like v1.1-pre or v1.1-alpha? Not sure how those things are handled in network protocols generally.)

ArthurSens (approved):
Let's do it :)

proposals/2024-01-29_native_histograms_text_format.md
fstab, line 51: I think it would be helpful to add these examples:

* Native histogram with labels.
* Native histogram that also has classic histogram buckets
* Native histogram with multiple Exemplars but without classic histogram buckets

fstab (dismissed):

proposals/2024-01-29_native_histograms_text_format.md
beorn7: Good catch. Native histograms can have repeated exemplars, and exemplars MUST have a timestamp, so the exemplar syntax is not trivially the same as for other samples.

csmarchbanks: Happy to add examples for the first two, multiple exemplars are trickier though. In the linked design doc the thinking was that adding multiple exemplars is out of scope for the initial work and should be a separate proposal. See https://docs.google.com/document/d/1qoHf24cKMpa1QHskIjgzyf3oFhIPvacyJj8Tbe6fIrY/edit?disco=AAABFtLHXX0. Do you think we need to figure out text format support for multiple exemplars first?

csmarchbanks: Ahh interesting, I didn't know that was a holdover from protobuf. Happy to use plural forms in this format, they certainly make more sense for a reader.

csmarchbanks: I have pushed a change adding multiple exemplar support as a non-goal, but I can figure something out if we decide it is an important goal to have as part of this work.

csmarchbanks:
> However, it might be good to sketch out a plan how to version the implementation as long as there is no official OM v1.1. (Something like v1.1-pre or v1.1-alpha? Not sure how those things are handled in network protocols generally.)

I added an option to the doc, happy to iterate on this now or later. It's always possible to amend these proposals as the work becomes more concrete.

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+Prometheus client libraries such as Python do not want to require a dependency on protobuf in order to expose native histograms, and in some languages protobuf is painful to use. Gating native histograms only to clients/users willing to use protobuf hurts adoption of native histograms, therefore, we would like a way to represent a native histogram in the text based format.
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Support native histograms in the text format
+* (Secondary) Encode/decode efficiency
+* (Secondary) Ease of implementation for client libraries
+* (Secondary) Human readibility of the format
```
fatsheep9146, line 34: small nit, readibility -> readability

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+hist_with_classic_buckets_sum 100
+```
+The metric will have no "magic" suffixes, then the value for each series is a custom struct format with the following fields:
+* `sum: float64` - The sum of all observations for this histogram. Could be negative in cases with negative observations.
+* `count: uint64` - The number of samples that were observed for this histogram.
+* `schema: int32` - The schema used for this histogram, currently supported values are -4 -> 8.
+* `zero_threshold: float64` - The width of the zero bucket.
+* `zero_count: uint64` - The number of observations inside the zero bucket.
+* `negative_spans: []BucketSpan` - The buckets corresponding to negative observations, optional.
+* `negative_deltas: []int64` - The delta of counts compared to the previous bucket.
```
fatsheep9146, line 71: Is this field optional too?

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+hist_with_labels{foo="bar",baz="qux"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+
+# TYPE hist_with_classic_buckets histogram
+hist_with_classic_buckets {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+hist_with_classic_buckets_bucket{le="0.001"} 4
+hist_with_classic_buckets_bucket{le="+Inf"} 24
+hist_with_classic_buckets_count 24
+hist_with_classic_buckets_sum 100
+```
+The metric will have no "magic" suffixes, then the value for each series is a custom struct format with the following fields:
```
krajorama, line 64: Not sure I follow this sentence. 
```suggestion
Series with native histogram values will have no "magic" suffixes and the value for each series is a custom struct format with the following fields inside curly braces:
```

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+Extend the OpenMetrics text format to allow structured values instead of only float values. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto). Starting with examples and then breaking up the format:
+```
+# TYPE nativehistogram histogram
+nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+
+# TYPE hist_with_labels histogram
+hist_with_labels{foo="bar",baz="qux"} {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
+
+# TYPE hist_with_classic_buckets histogram
+hist_with_classic_buckets {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
```
krajorama, line 58: Let's define somewhere that for any combination of label values, the native histogram series comes first and the classic series come after it. (This will make sure that our current scrape implementations work without much rework. See https://github.com/prometheus/prometheus/blob/bfdca40fd2fa78c7d80f335c0b67cdaa2b5432fd/model/textparse/protobufparse.go#L58 . Also it's better to be explicit.)

And from that it follows that when there are multiple label value combinations, you always get native, classic, native, classic, etc

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+
+* Requiring backwards compatability (OpenMetrics 2.0 would be ok), and especially forwards compatability (not required in the OpenMetrics spec).
+* Support for multiple exemplars (this will be done in a future proposal).
+
+## How
+
+Extend the OpenMetrics text format to allow structured values instead of only float values. This structured value will be used to encode a structure with the same fields as is exposed using the [protobuf exposition format](https://github.com/prometheus/client_model/blob/master/io/prometheus/client/metrics.proto). Starting with examples and then breaking up the format:
+```
+# TYPE nativehistogram histogram
+nativehistogram {count:24,sum:100,schema:0,zero_threshold:0.001,zero_count:4,positive_spans:[0:2,1:2],negative_spans:[0:2,1:2],positive_deltas:[2,1,-2,3],negative_deltas:[2,1,-2,3]}
```
krajorama, line 52: I wonder if we should explicitly say that new line is not allowed inside the structured part.

proposals/2024-01-29_native_histograms_text_format.md
krajorama: Might be worth adding a simple example with two sets of labels, something like id="1"  and another set of series with id="2".

krajorama (commented):
Looks good, easy to convert to internal data. I added some minor comments.

proposals/2024-01-29_native_histograms_text_format.md
beorn7: By now, we have added multiple exemplars to the protobuf spec. Which means it's now way more interesting to do the same for the upcoming text format. However, it is not strictly required, either. If you think it complicates things significantly, you could prefer an incremental approach.

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+Note that the goals of efficiency and human readability are commonly at odds with each other.
+
+### Audience
+
+Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
+
+## Non-Goals
+
+* Requiring backwards compatability (OpenMetrics 2.0 would be ok), and especially forwards compatability (not required in the OpenMetrics spec).
+* Support for multiple exemplars (this will be done in a future proposal).
```
bwplotka, line 45: Can this be done in non-breaking change?

proposals/2024-01-29_native_histograms_text_format.md
bwplotka, line 64:
```suggestion

Native histograms will reuse "histogram" type. It can also reuse the same name family metric name (when clients decided to expose histogram as classic and native one in the same app). 

When a histogram metric series under this type have any of the "magic" histogram suffix e.g. `_bucket`, `_sum` and `_count`, parsers should expect float values. Otherwise, the value for each series is a custom struct format with the following fields:
```

proposals/2024-01-29_native_histograms_text_format.md
bwplotka: +1, mentioning which ones are optional would be nice

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Pitfalls of the current solution
+
+Prometheus client libraries such as Python do not want to require a dependency on protobuf in order to expose native histograms, and in some languages protobuf is painful to use. Gating native histograms only to clients/users willing to use protobuf hurts adoption of native histograms, therefore, we would like a way to represent a native histogram in the text based format.
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Support native histograms in the text format
```
bwplotka, line 31: Is support of custom (explicit) bucketing a goal of this proposal? (I hope so! 🙈 )

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+* (Secondary) Ease of implementation for client libraries
+* (Secondary) Human readibility of the format
+
+Note that the goals of efficiency and human readability are commonly at odds with each other.
+
+### Audience
+
+Client library maintainers, OpenMetrics, and Prometheus scrape maintainers.
+
+## Non-Goals
```
bwplotka, line 42: Also a non goal (I guess): Created timestamp on native histograms

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Backwards compatibility and semantic versioning
+
+After discussions with a few people it is believed that these changes can be made in a 1.x release of OpenMetrics. OpenMetrics 1.x parsers that support native histograms will still be able to read OpenMetrics 1.0 responses, therefore this change is backwards compatible. However, this change is not forwards compatible, i.e. an OpenMetrics 1.0 parser will not be able to read an OpenMetrics >= 1.1 response. Any producers implementing native histograms MUST also implement content negotiation and fall back to OpenMetrics 1.0.0, and therefore not expose native histograms, if a supported version cannot be negotiated. Note that the behavior to fall back to 1.0.0 is already part of the [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#protocol-negotiation).
```
bwplotka, line 83: > After discussions with a few people it is believed that these changes can be made in a 1.x release of OpenMetrics. OpenMetrics 1.x parsers that support native histograms will still be able to read OpenMetrics 1.0 responses, therefore this change is backwards compatible. However, this change is not forwards compatible, i.e. an OpenMetrics 1.0 parser will not be able to read an OpenMetrics >= 1.1 response

Hm, I think this text misunderstands backward compatibility. To me, if 1.0 parser fails to read 1.x format (instead of e.g. skipping native histograms), that means native histograms in 1.x format is NOT backward compatible change. I would say let's be explicit about this. This change is NOT backward compatible, so the content negotiation mechanism HAS to be added to release it in 1.x (and that's fine). 

```suggestion
Implementing this change in 1.x would be considered a breaking change, because it could break 1.0 parsers. For this reasons this feature has to be behind a content negotiation. Any producers implementing native histograms MUST also implement content negotiation and fall back to OpenMetrics 1.0.0, and therefore not expose native histograms, if a supported version cannot be negotiated. Note that the behavior to fall back to 1.0.0 is already part of the [OpenMetrics spec](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#protocol-negotiation).
```

proposals/2024-01-29_native_histograms_text_format.md
```diff
@@ -0,0 +1,110 @@
...
+
+Until a version of OpenMetrics is released that contains a stable version of native histograms consumers can determine if native histograms may be present by asking for a `nativehistogram` pre-release identifier. For example,
+```
+Accept: application/openmetrics-text;version=1.1.0-nativehistogram.*,application/openmetrics-text;version=1.0.0,text/plain;version=0.0.4
+```
+would mean the consumer can accept any nativehistogram enabled pre-release version of OpenMetrics 1.1.0, the base 1.0.0 version of OpenMetrics, or the 0.0.4 version of the classic Prometheus text format. Producers must include the proper content type for their version, such as the first nativehistogram pre-release:
+```
+Content-Type: application/openmetrics-text;version=1.1.0-nativehistogram.0
+```
```
bwplotka, line 92: Hm, this is interesting. What's "release identifier? I assume you want to version SPEC and native histogram feature separately? Why? Can we make it simpler?

I think I would expect something like:

```
Accept: application/openmetrics-text;version=1.1.0;feature=nativehistogram,application/openmetrics-text;version=1.0.0,text/plain;version=0.0.4
```

.. then we could simply do 1.1.1 or 1.2.0 if you have significant histogram change. Let's not create sub versions of every feature, this will make it bit painful to use and maintain. Speaking from experience when we evaled different options for remote write v2: https://github.com/prometheus/docs/pull/2462

proposals/2024-01-29_native_histograms_text_format.md
bwplotka, line 52: Can we establish some simple forward compatibility rules for this JSON e.g. Parsers MUST ignore unknown fields? Would help in future.

bwplotka (approved):
Overall LGTM. Added some suggestions though. Good work!

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Oh, interesting idea, it would make parsing easier, but I don't know that anything in OpenMetrics requires order of series today so I am not sure if we want to be that strict or not...

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I am actually going to take this one further and propose that white space is not allowed inside the structure. That allows parsers to separate the labels, value, and timestamp all with spaces.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I am neutral on support for created timestamps, seems unlikely too many people will use them, but:
```
my_histogram {...value...} 1717524294.000
```
would work with this proposal.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: See https://semver.org/#spec-item-9, it is just a pre-release, IMO if `version=1.1.0` is there without any sort of prerelease suffix native histograms must be fully supported and there shouldn't be extra feature tags. I would also be ok with something like `alpha.0`, `alpha.1`, etc... if we don't want to specifically include nativehistogram in the version.

proposals/2024-01-29_native_histograms_text_format.md
ArthurSens: I wouldn't change how created timestamps are represented in OM in this proposal. We sure want to change it, but it would probably be better to change created timestamps in all metric types at once instead of changing for native histograms and others later.

ArthurSens: FYI: [here is how created timestamps are represented for histograms in OM today](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#histogram-1:~:text=in%20any%20order.-,Histogram,appear%20in%20a%20Sample%20with%20a%20MetricName%20with%20the%20suffix%20%22_count%22.,-Buckets%20MUST%20be).

For native histograms I think we could to the same for now? Just add a new line, repeating the metric name and using the suffix `_created`

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I disagree, I think the behavior you describe is [forward compatibility](https://en.wikipedia.org/wiki/Forward_compatibility). Input designed for a 1.1 parser does not need to be handled by a 1.0 parser to classify as backwards compatible, the OpenMetrics spec makes sure of this by requiring all producers to fall back to 1.0.0 if they are not explicitly told otherwise.

Maybe this is semantics as to what a "breaking change" is though in the OpenMetrics world. If I see text such as "Implementing this change in 1.x would be considered a breaking change" I would expect this to require a major version bump which I don't think is what we want, and would also mean it is impossible to add anything to OpenMetrics 1.x as the format is incredibly locked down.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I would be ok with that, but I am also not sure it is necessary with content negotiation as new fields would require part of a new version of OpenMetrics. I.e. I am unsure if we want to impose the level of strictness a given parser needs to implement, and it might be ok for different parsers to make different choices.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Correct me if I have missed something, but I think the current proposal for that is starting with classic histograms and storing them as a custom native histogram internally? If so this proposal supports that. This proposal does not support defining custom buckets outside of the classic histogram format, but could be done fairly trivially, and alongside any protobuf changes, in OpenMetrics 1.2.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Yes (by my definition of non-breaking at least, e.g. OpenMetrics 1.2), there are a couple ideas of how to do this, we could provide a list of `# {labels} <value> <timestamp> # {labels} <value> <timestamp>` as exemplars at the end of the line instead of just one, or add an `exemplars:[]` list to the structure proposed in this proposal.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Ahh, missed the "Created" part of the original comment. I agree that we will not change how created timestamps are represented in OpenMetrics today. We will continue to add an additional `_created` when native histograms are in use.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I updated the language with this and the other suggestion here. Hopefully that helps clarify how to tell native histogram and classic histogram series apart.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Yeah, I want to keep the discussion separate. I think there are valid use cases for multiple exemplars on non-native histogram series as well, and don't want to bring that discussion into this format which is mostly agreed upon. If they both end up in OpenMetrics 1.1 that would be great though!

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Yep, all of the lists are optional, updated!

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Both you and bartek had suggestions for this sentence, changed them around some, let me know if that is better!

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I guess I have been thinking about it similar to Go compiler version changes. I don't expect code written for Go 1.22 to work with Go 1.0, but I do expect code written for Go 1.0 to still work with Go 1.22. I think of Go as backwards compatible so also think that this should be backwards compatible.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: It looks like histogram `le` buckets already MUST be in order so it seems reasonable to at least have "native histograms SHOULD come before classic histogram series for each labelset" or something to that effect. I have added a bit to require native histograms first along with an example with two labelsets.

proposals/2024-01-29_native_histograms_text_format.md
fstab: I think we need to put some thoughts into versioning. The current proposal does not cover all of native histograms, Exemplars, created timestamps are out-of-scope.

Now, how will content negotiation work if we add these in the future? How will a client library know whether to include a list of exemplars in native histogram text format or not?

proposals/2024-01-29_native_histograms_text_format.md
krajorama: Yes, current proposal for custom buckets is to upconvert on scrape. So it's super useful for legacy apps where you don't have the code/compiler anymore.

beorn7: Indeed. I wouldn't see custom buckets directly in the text format a strict necessity. On the contrary, the classic representation is way more readable, and it will result in the same NHCB upon ingestion.

proposals/2024-01-29_native_histograms_text_format.md
beorn7: I guess just repeating the exemplar as we know it should work just fine. Maybe I'm missing something, but it feels so easy that we might just include it right now…

proposals/2024-01-29_native_histograms_text_format.md
beorn7: BTW: It's not JSON.

beorn7: While I like a way of adding stuff that old parsers will just ignore, this concept doesn't exist in OM so far, so it might be weird to introduce it in a sub-set of the protocol. Let's maybe solve that more generally in OM v2.

proposals/2024-01-29_native_histograms_text_format.md
beorn7: Hehehe, @bwplotka and I nerded out about this on his podcast. I also find it hard to clearly define what is forward and backward compatibility. I also understood OM v1 formulated in a way that there isn't much you could change in a v1.1. However, the view expressed in this design doc makes way more sense to me. And reading the section about content negotiation in OM from this perspective suddenly also felt like it was meant that way by OM. However, discussions in the past lead to inconclusive results about this.

I would like most if we adopted the view expressed in the design doc, accepting that the current OM spec can be interpreted this way, but we will also clarify the intent in OM 1.1.

proposals/2024-01-29_native_histograms_text_format.md
beorn7: I would opt for some way of opting into the feature as an experimental one, without any changes to the standard yet. Based on the experience made, we can make the cut what should land in v1.1 later (including maybe other changes being discussed). In other words: I don't think this design doc has to exactly describe v1.1 already.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Does it sound good to optimistically move forward with the view in this design doc, and if we receive pushback when creating the OM 1.1 changes we can find new consensus then if required?

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I also think the repeating exemplar should work just fine, I added a paragraph to support that format as well. There is a bit of an open question around whether we support a list of exemplars on all series or not, the current proto proposal states "Only used for native histograms", I would like to see them on all series, but we can also figure that out when writing the final OpenMetrics 1.1 spec.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: I am happy to implement some other versioning scheme, the important part to me is that any exposers add some sort of pre-release identifier until OM 1.1 is actually released, and force ingesters to specify that they are ok with a pre-release version. A features section alongside a pre-release version seems reasonable to me, then becomes unnecessary once 1.1 is released. That allows us some flexibility to make breaking changes to the unreleased features. E.g. if we change around where the list of exemplars are stored, or choose to use a less compact format after feedback.

Content negotiation for features added in 1.2, 1.3, etc... will require client libraries to keep track of the features added in each minor version and leave them out if the requested version is lower. That could mean fall all the way back to 1.0, or maybe have support for a few versions (1.0, 1.1, 1.4) and choose the highest version less than or equal to what is requested by the ingester. E.g. if Prometheus requests 1.3, the client would return 1.1 in that example. 1.0 is the only version that all producers MUST support.

Also note that created timestamps have always been in scope (no change from current text format), and I just added a proposal for exemplar support.

proposals/2024-01-29_native_histograms_text_format.md
bwplotka: Nice, my point is that to ensure it **can** be done in non breaking change, then it can be postponed (:

proposals/2024-01-29_native_histograms_text_format.md
bwplotka: Shameless link to the mentioned episode with compatibility discussions: https://open.spotify.com/episode/2TPgtGv3boEegKQI9nTeE5?si=rMRhzSZ1T2aL_kjULPTZeA 🙈

proposals/2024-01-29_native_histograms_text_format.md
bwplotka: I don't think I fully agree, there are benefits to nhcb on the text exposition format:

* nhcb would make SDKs overhead smaller (e.g. see [Otel Java benchmarks](https://opentelemetry.io/blog/2024/java-metric-systems-compared/) on histograms - while the whole benchmark could be argued on, the classic histograms have some overhead as they are treated as separate lines on scrape)
* nhcb would make server scrape cheaper (probably not significant though?)

But maybe you argue proto should be default for performance? But it was just said that some languages would never do proto..

Anyway, I am happy to scope this down here, do you mind @csmarchbanks to add nhcb as explicit No-Goal item? Thanks! 🤗

proposals/2024-01-29_native_histograms_text_format.md
bwplotka: Do you mind adding this to non goal? e.g. `* Changing created timestamp semantics (it will be a separate series as in OM 1.0)`

proposals/2024-01-29_native_histograms_text_format.md
bwplotka: Cool, thanks. Let's make it clear it's not a JSON somewhere please 🙈

proposals/2024-01-29_native_histograms_text_format.md
bwplotka: Thanks! After reading all here, I still think my suggestion makes sense. First and foremost, because I drop mention what's backward and what's forward, so we should all agree on this. (: Do you mind double-checking?

We can argue on what's backward, what's forward in separate place ;p

bwplotka: In separation to my suggestion, in terms of what backward what's forward, I spent some time thinking about it and it feels it depends on the **point of view.**, or what is the **subject**.

### Details

Sounds like "the view expressed by this design doc is":

* forward compatibility = OM 1.0 parser works with OM 1.1
* backward compatibility = OM 1.x parser works with OM 1.0

If we replace C as client and P as protocol:

* forward: old C <-> new P 
* backward: new C <-> old P

This could make sense if we version parsers here (clients), so let's call it "Client View". But do we talk about parser compatibility or protocol in this section?

To me, in this context, it's opposite, and it's also how we formulated remote write spec compatibility e.g. [here](https://deploy-preview-2462--prometheus-docs.netlify.app/docs/specs/remote_write_spec_2_0/#backward-and-forward-compatibility). We put protocol in the center: Backward compatibility exists if new spec changes (e.g. OM 1.1) can be read by OM 1.0 client/parser without trouble (while skipping newest features). Forward compatibility in my mind is OM 1.1 client reading OM 1.0 which is easier to maintain, thus rarely mentioned in practice. Let's call this "Protocol View".

In our new convention, the Protocol View means:

* forward: old P <-> new C
* backward: new P <-> old C

This is perhaps also what SemVer assumes by saying [`MINOR version when you add functionality in a backward compatible manner`](https://semver.org/) in my mind. This means old C will work with new P (e.g. new dep). The new C working with old P (e.g. dep) (forward compatibility here?) does not make sense in this "upgrade" context.

To me, it seems like generally it's used both ways and it depends on the point of view e.g. do you see it from the point of view of the protocol or API ("new protocol" can be consumed by "old client"), or client ("new client" can consume "old protocol"). E.g. check [the wikipedia definition](https://en.wikipedia.org/wiki/Backward_compatibility)

> In [telecommunications](https://en.wikipedia.org/wiki/Telecommunications) and [computing](https://en.wikipedia.org/wiki/Computing), backward compatibility (or backwards compatibility) is a property of an [operating system](https://en.wikipedia.org/wiki/Operating_system), software, real-world product, or [technology](https://en.wikipedia.org/wiki/Technology) that allows for [interoperability](https://en.wikipedia.org/wiki/Interoperability) with an older [legacy system](https://en.wikipedia.org/wiki/Legacy_system), or with [input](https://en.wikipedia.org/wiki/Input/output) designed for such a system.

You can say both:

A) Backward compatibility is when "new client works for older API/protocol", so new C <-> old P
B) Backward compatibility is when "new API/protocol works for older client", so new P <-> old C

Both match definition kind of? This is similar in Go:

> I don't expect code written for Go 1.22 to work with Go 1.0, but I do expect code written for Go 1.0 to still work with Go 1.22. I think of Go as backwards compatible so also think that this should be backwards compatible.

First of all, funny enough the https://go.dev/doc/go1compat does not mention "backward" or "forward" prefix anywhere 🙈 But indeed everyone says "Go is backward compatible".

Secondly you assume `code written for Go 1.0` is `old C` and Go 1.22 SDK is `new P` and you expect that to work, so `old C <-> new P` is expected, and what's expected =  backward compatibility, which supports "protocol view", no?

In other words, to clone this to OM, it's backward compatibility when old parser (aka code written for Go 1.0) works with new protocol (Go 1.22 SDK)? This proves that one can assume client vs protocol as opposite things, and semantically it can still work.

## Summary

Both definition can technically work, but in my mind we change (and version!) protocol here, so P means protocol, not client here. It's a breaking change if old C does not work with new P with minor version bump (aka OM 1.0 parser with OM 1.1 spec).

From this POV, forward compatibility is when new C that supports new P cannot possibly work with old P, which is .. hard to even achieve in this context?

Interestingly enough we switch the definition if we talk about [senders "being backward" compatible with old receivers (via old protocol)](https://deploy-preview-2462--prometheus-docs.netlify.app/docs/specs/remote_write_spec_2_0/#backward-and-forward-compatibility:~:text=When%20talking%20to%201.x%20Receiver%2C%20Sender%20SHOULD%20use%20Content%2DType%3A%20application/x%2Dprotobuf%20for%20backward%20compatibility.), which proves both direction can be useful and can be mixed... as long as you are explicit/clear what is your main subject.

proposals/2024-01-29_native_histograms_text_format.md
beorn7: In the old days, we usually said "text format is for simple cases, if you want industry strength performance, use protobuf" (which ironically didn't really work out well back then because the Go protobuf implementation was so wasteful).

But I think this argument still makes some sense. You could say that languages that just don't want protobuf at all are probably also the languages you wouldn't use if performance is a core concern (Python being a good example for this case).

proposals/2024-01-29_native_histograms_text_format.md
beorn7: Thanks for the interesting thoughts. It feels to me like actual experts on wire formats should have already found a clear definition for all of this. Do we have such an expert among us? It would be best to go with the view that already is a consensus among experts rather than coming up with something on our own (even though it might be self-consistent).

My preliminary conclusion is that OpenMetrics _might_ have in mind what @csmarchbanks explained in more detail in this PR. 
Of course, we could also define that an old OM parser must be able to parse any OM 1.x protocol, but given the terse way OM 1.0 is defined, I doubt we could ever have a 1.1. Which is another hint to me that OM always had @csmarchbanks's view in mind (but I agree that it wasn't clearly explained because I assumed initially that OM 1.1 is essentially impossible).

csmarchbanks: @RichiH or @SuperQ perhaps one of you could weigh in on the intention of OpenMetrics versioning?

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Any suggestions as to how to make it more clear that it is not JSON? In the format definition I have "The value for each series of a native histogram is a custom struct format" and I specifically compare it to JSON with the phrase "Compared to JSON this avoids consistently repeating keys and curly braces" after the type definitions.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Done!

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: Added as a non-goal as requested. Specifically:
```
Support for custom buckets as part of the native histogram structure. If needed by a consumer, custom buckets will need to be parsed from classic buckets.
```
to make it clear how consumers can use custom buckets if desired.

proposals/2024-01-29_native_histograms_text_format.md
csmarchbanks: My concern with your suggestion is that writing that this proposal is a breaking change makes me think it should be in 2.0 regardless of content negotiation (content negotiation doesn't allow us to break semver). The crux of the issue for me is whether we can support this feature in 1.1 or are forced to do a major version bump. If we think of this change as breaking then it should be put into 2.0 IMO. Following onto that, if we think this needs a major version bump we will need significantly more care in what else we include with a release, and think more about how to allow future changes without another major version bump (e.g. adding in a way to add new fields like what protobuf has, but does not exist in OM 1.0).

If we (and the OpenMetrics team) can agree that this can be released in 1.x, I think we could find some phrasing that works for both of us. I am not attached to the backwards/forwards compatibility wording, and agree that it depends on the point of view of the producers/consumers (in OpenMetrics parlance) vs of the protocol itself.

As @beorn7 mentions I think the only way we can release this (and any other new feature) in OM 1.x is if we take the client viewpoint. The client viewpoint is also what allows version content negotiation to be effective as content negotiation (for a version, clearly content negotiation is useful to negotiate text vs protobuf) is not helpful if all clients MUST be able to read all versions of a protocol anyway. This is the reason I think(hope?) the client view was possibly taken by the OpenMetrics team, but let's wait for some of them to respond :).

> Secondly you assume `code written for Go 1.0` is `old C` and Go 1.22 SDK is `new P` and you expect that to work, so `old C <-> new P` is expected, and what's expected = backward compatibility, which supports "protocol view", no?

Other way around is how I was thinking about it, the code written for Go 1.0 would be `old P` (the code is the text format protocol in this comparison), Go 1.22 compiler is `new C` as it is the new client consuming the protocol. Therefore `new C <-> old P`.

csmarchbanks:
I have added this as an agenda item for a dev summit. It seems like there are a couple to be decided points still, mainly around minor or major version bump. Hopefully having the sync discussion will allow us to come to a conclusion.

brancz:
We had a few consensus items at the dev summit that I think will allow this to move forward:

* We intend to add native histograms as a part of OpenMetrics 1.1. A major version bump is not required.
* We agree that the understanding of backwards compatibility is that a A.y parser can also parse an A.x format. But the A.x parser doesn’t need to parse the A.y format. Where y > x, A, x, y being integers
* OpenMetrics doesn’t need a major version bump if the changes are backwards compatible.

Looking forward to seeing this become a reality!

csmarchbanks:
🎉 

With the consensus items from the dev summit, are there any objections to merging this as is? We can continue to tweak it as development work proceeds. If I don't hear any objections I will plan to merge sometime next week.

beorn7:
I think the only somewhat open point is how to mark pre-release features in content negotiation. At the same time, I think that's an aspect where we don't have to be terribly descriptive and precise in the first place, because it is just for pre-releases and precisely not for the stable standard. @csmarchbanks rightfully pointed to https://semver.org/#spec-item-9 , which essentially allow almost anything after the hyphen, and what's proposed here in the doc is semver compliant and also makes sense to me. As a design doc is not describing the final implementation, we have some wiggle room here, so I would propose to merge this as-is and see how it shakes out in the implementation.

beorn7 (approved):

beorn7:
I don't think we should change the Prometheus text format, if we can avoid it.

As we concluded above, this could be part of OM 1.1. If it is part of OM 1.1, it should also be part of OM 2.0. If OM 2.0 happens soon, we might never have an OM 1.1, then this would only go into OM 2.0.

bwplotka:
FYI: https://github.com/prometheus/OpenMetrics/issues/279 - I propose this should be in 2.0, no 1.1 is planned.

1atAlcone (approved):

PR #35, opened 2024-04-09
remote write 2.0 proposal

alexgreenbank:
Draft proposal for the ongoing Remote Write 2.0 project.

Mostly a pointer to the draft spec doc.

PR #36, opened 2024-05-08
Updated proposal: new structure, all decisions outlined.

bwplotka:
Starting point, some TBDs.


bwplotka:
Let's merge into https://github.com/prometheus/proposals/pull/35 and iterate there 🤗

PR #37, opened 2024-05-14
Add proposal: Simplify joins with info metrics in PromQL

aknuds1:
I'm adding a proposal about enhancing Prometheus with a 1st class feature to PromQL for easy joining with info type metrics. This is in particular supposed to facilitate OpenTelemetry resource attributes (via the `target_info` metric they get mapped to), but works for info metrics in general.

Please note that in agreement with @codesome, I'm deliberately punting on TSDB details, as @codesome wants to have a separate proposal for those.

### DIFF
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -150,5 +150,5 @@ Instead of encoding metadata, e.g. OTel resource attributes, as info metric labe
 
 The tasks to do in order to migrate to the new idea.
 
-* [ ] Task one <GH issue>
-* [ ] Task two <GH issue> ...
+* [ ] [Add experimental PromQL `info` function MVP](https://github.com/prometheus/prometheus/pull/14495)
+* [ ] Extend `info` function MVP with the ability to support `info` metrics in general, with persistence of info metrics metadata
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -65,7 +65,22 @@ Goals and use cases for the solution as proposed in [How](#how):
 * Track when info metrics' set of identifying labels changes. This shouldn't be a frequent occurrence, but it should be handled.
 * Automatically treat the old version of an info metric as stale for query result enriching purposes, when its data labels change (producing a new time series, but with same identity).
 * Add TSDB API for, given a certain time series and a certain timestamp, getting data labels, potentially filtered by certain matchers, from info metrics with identifying labels in common with the time series in question.
-* Simplify enriching of query results with info metric labels in PromQL, e.g. via a new function.
+  * If no data label matchers are provided, _all_ the data labels of found info metrics are added to the resulting time series.
+  * If data label matchers are provided, only info metrics with matching data labels are considered.
+  * If data label matchers are provided, _precisely_ the data labels specified by the label matchers are added to the returned time series.
+  * If data label matchers are provided, time series are only included in the result if matching data labels from info metrics were found.
+  * A data label matcher like `k8s_cluster_name=~".+"` guarantees that each returned time series has a non-empty `k8s_cluster_name` label, implying that time series for which no matching info metrics have a data label named `k8s_cluster_name` (including the case where no matching info metric exists at all) will be excluded from the result.
+  * A special case: If a data label matcher allows empty labels (equivalent to missing labels, e.g. `k8s_cluster_name=~".*"`), it will not exclude time series from the result even if there's no matching info metric.
+  * A data label matcher like `__name__="target_info"` can be used to restrict the info metrics used.
+    However, the `__name__` label itself will not be copied.
+  * Label collisions: The input instant vector could already contain labels that are also part of the data labels of a matching info metric.
+    Furthermore, since the info function might find multiple differently named info metrics with matching identifying labels, those might have overlapping data labels.
+    In this case, the info function has to check if the values of the affected labels match or are different.
+    The former case is not really a label collision and therefore causes no problem.
+    In the latter case, however, the function has to return an error.
+    The collision can be resolved by constraining the labels via the optional label-selector argument of the info function.
+    And of course, the user always has the option to go back to the original join syntax (or, even better, avoiding ingesting conflicting info metrics in the first place).
+* Simplify enriching of query results with info metric data (non-identifying) labels in PromQL, e.g. via a new function, based on aforementioned TSDB API.
 
 ### Audience
 
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -1,0 +1,78 @@
+## Native Support for Info Metrics Metadata
+
+* **Owners:**
+  * Arve Knudsen [@aknuds1](https://github.com/aknuds1) [arve.knudsen@grafana.com](mailto:arve.knudsen@grafana.com)
+
+* **Implementation Status:** Partially implemented
+
+* **Related Issues and PRs:**
+  * [WIP: Info PromQL function prototype](https://github.com/grafana/mimir-prometheus/pull/598)
+
+* **Other docs or links:**
+  * [Proper support for OTEL resource attributes](https://docs.google.com/document/d/1FgHxOzCQ1Rom-PjHXsgujK8x5Xx3GTiwyG__U3Gd9Tw/edit#heading=h.unv3m5m27vuc)
+  * [Special treatment of info metrics in Prometheus](https://docs.google.com/document/d/1ebhGNLs3uhdeprJCullM-ywA9iMRDg_mmnuFAQCloqY/edit#heading=h.2rmzk7oo6tu8)
+
+> This proposal collects the requirements and implementation proposals for enhancing Prometheus with native support for info metrics metadata.
+
+## Why
+
+Currently Prometheus "forgets" which are the identifying labels of info metrics upon ingestion, even though this information is present in at least the OpenMetrics protobuf exposition format (the OpenMetrics text exposition format unfortunately lacks this capability).
+The fact that Prometheus lacks a notion of which are info metrics' identifying labels leads to certain problems:
+
+* Explicit knowledge of each info metric's identifying labels must be embedded in join queries for when you wish to enrich queries with data (non-identifying) labels from info metrics.
+* Complex join queries must be written in order to enrich time series with corresponding labels from info metrics.
+  This is particularly problematic in the OpenTelemetry (AKA OTel) context, since users depend on (joining with) the `target_info` info metric in order to add relevant [resource attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md) back to their Prometheus metrics.
+* If an info metric's data (non-identifying) labels change (a situation that should become more frequent with OTel in the future, as the model will probably start allowing for non-identifying resource attribute mutations), join queries against the info metric (e.g. `target_info`) will temporarily fail due to resolving the join keys to two different metrics, until the old metric is marked stale (by default after five minutes).
+
+Especially in order to provide the best possible OTel experience, the info metric (`target_info` in the case of OTel) staleness problem needs to be solved, so users won't experience temporarily failing join queries while trying to include OTel resource attributes.
+Also, it would be much better if we could provide a simpler query experience where the user doesn't have to know how to write PromQL joins (a fairly complex matter), in order to include e.g. OTel resource attributes.
+Another possible positive outcome might be dedicated support in the Grafana UI for visualizing the resource attributes of each OTel metric.
+
+### Pitfalls of the current solution
+
+Prometheus currently persists info metrics as if they were normal float samples.
+This means that knowledge of info metrics' identifying labels are lost, and you have to base yourself on convention when querying on them (for example that `target_info` should have `job` and `instance` as identifying labels).
+The consequence is that you need relatively expert level PromQL knowledge to include info metric labels in your query results; as OTel grows in popularity, this becomes more and more of a problem as users will want to include certain labels from `target_info` (corresponding to OTel resource attributes).
+Without persisted info metric metadata, one can't build more user friendly abstractions (e.g. a PromQL function) for including OTel resource attributes (or other info metric labels) in query results. Neither can you build dedicated UI for OTel resource attributes (or other info metric labels).
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Persist info metrics with known identifying labels as a new info metric sample type.
+* Store for each info metric sample (of the new type) which are the identifying labels.
+* Store in the TSDB immediately that the previous version of an info metric is stale, when its data labels change.
+* Add TSDB API for, given a certain time series and a certain timestamp, getting data labels, potentially filtered by certain matchers, from info metrics with identifying labels in common with the time series in question.
+* Simplify inclusion of info metric labels in PromQL.
+
+### Audience
+
+Prometheus maintainers.
+
+## Non-Goals
+
+## How
+
+* A new info metric sample type will be introduced, where the sample value is the info metric's identifying labels.
+* The head and block indexes will be augmented with indexes of info metrics.
+* A method will be added to the TSDB API for matching info metric data labels to a time series, given a certain timestamp and potentially data label matchers - the method will use the aforementioned head and block info metric indexes.
+* Thanks to the head and block info metric indexes, the info metric staleness problem should be solved, since one can pick the latest version of the info metric for overlapping time ranges.
+* We propose simplifying the inclusion of info metric labels in PromQL through a new `info` function (TODO: describe).
+
+* Make it concise and **simple**; put diagrams; be concrete, avoid using “really”, “amazing” and “great” (:
+* How you will test and verify?
+* How you will migrate users, without downtime. How we solve incompatibilities?
+* What open questions are left? (“Known unknowns”)
+
+## Alternatives
+
+The section stating potential alternatives. Highlight the objections reader should have towards your proposal as they read it. Tell them why you still think you should take this path [[ref](https://twitter.com/whereistanya/status/1353853753439490049)]
+
+1. This is why not solution Z...
+
+## Action Plan
+
+The tasks to do in order to migrate to the new idea.
+
+* [ ] Task one <GH issue>
+* [ ] Task two <GH issue> ...
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -120,6 +120,7 @@ sum by (k8s_cluster_name, http_status_code) (
 TODO:
 
 * Specify detection of info metric identifying labels for other ingestion methods than OTLP.
+* Define how this functionality would work together with OOO samples.
 
 ## Alternatives
 
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -124,11 +124,6 @@ sum by (k8s_cluster_name, http_status_code) (
 )
 ```
 
-TODO:
-
-* Specify detection of info metric identifying labels for other ingestion methods than OTLP.
-* Define how this functionality would work together with OOO samples.
-
 ## Alternatives
 
 ### Add metadata as prefixed labels
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -83,6 +83,8 @@ Goals and use cases for the solution as proposed in [How](#how):
     The collision can be resolved by constraining the labels via data label matchers.
     And of course, the user always has the option to go back to the original join syntax (or, even better, avoiding ingesting conflicting info metrics in the first place).
 * Simplify enriching of query results with info metric data (non-identifying) labels in PromQL, e.g. via a new function, based on aforementioned TSDB API.
+* Ensure backwards compatibility with current Prometheus usage.
+* Minimize potential conflicts with existing metric labels.
 
 ### Audience
 
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -146,6 +146,18 @@ Instead of encoding metadata, e.g. OTel resource attributes, as info metric labe
 * Increased series churn when metadata labels change
 * More labels per metric increases CPU/memory usage
 
+### Make the `info` function require specifying the info metric(s)
+
+Instead of letting `info` by default join with all matching info metrics, have it require specifying the info metric name(s).
+
+#### Pros
+
+* The user won't be confused about data labels being included from info metrics they didn't expect
+
+#### Cons
+
+* The UX becomes more complex, as the user is required to specify which info metric(s) to join with
+
 ## Action Plan
 
 The tasks to do in order to migrate to the new idea.
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -1,4 +1,4 @@
-# Add 1st class feature to PromQL for handling info type metrics
+# Simplify joins with info metrics in PromQL
 
 * **Owners:**
   * Arve Knudsen [@aknuds1](https://github.com/aknuds1) [arve.knudsen@grafana.com](mailto:arve.knudsen@grafana.com)
@@ -13,7 +13,7 @@
   * [Special treatment of info metrics in Prometheus](https://docs.google.com/document/d/1ebhGNLs3uhdeprJCullM-ywA9iMRDg_mmnuFAQCloqY/edit#heading=h.2rmzk7oo6tu8)
   * [Scenarios scratch pad](https://docs.google.com/document/d/1nV6N3pDfvZhmG2658huNbFSkz2rsM6SpkHabp9VVpw0/edit#heading=h.luf3yapzr29e)
 
-> This proposal collects the requirements and implementation proposals for adding a 1st class feature to PromQL for handling info type metrics.
+> This proposal collects the requirements and implementation proposals for simplifying joins with info type metrics in PromQL.
 
 ## Why
 
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -17,6 +17,11 @@
 
 ## Why
 
+Info metrics are [defined by the OpenMetrics specification](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info) as "used to expose textual information which SHOULD NOT change during process lifetime".
+Furthermore the OpenMetrics specification states that info metrics ["MUST have the suffix `_info`"](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info-1).
+Despite the latter OpenMetrics requirement, there are metrics with the info metric usage pattern that don't have the `_info` suffix, e.g. `kube_pod_labels`.
+In this proposal, we shall include the latter in the definition of info metrics.
+
 Currently, enriching Prometheus query results with corresponding labels from info metrics is challenging.
 More specifically, it requires writing advanced PromQL to join with the info metric in question.
 Take as an example querying HTTP request rates per K8s cluster and status code, while having to join with the `target_info` metric to obtain the `k8s_cluster_name` label:
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -50,6 +50,7 @@ There are other problems with Prometheus' current method of including info metri
 If Prometheus could persist info metrics' identifying labels (e.g. `job` and `instance` for `target_info`), human knowledge of the correct identifying labels may become unnecessary when "joining" with info metrics.
 Information about info metric identifying labels is present in at least the OpenMetrics protobuf exposition format (the OpenMetrics text exposition format unfortunately lacks this capability).
 It can also easily be deduced when ingesting metrics from OTLP (OTel Protocol).
+Most info metrics' identifying labels will be `job` and `instance`, but there are some exceptions (e.g. `kube_pod_labels`).
 Intrinsic knowledge of info metrics' identifying labels could also help in solving temporary conflicts between old and new versions of info metrics, when data (non-identifying) labels change.
 Another possible positive outcome might be dedicated support in UIs (e.g. Grafana) for visualizing the resource attributes of OTel metrics.
 
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -1,4 +1,4 @@
-## Native Support for Info Metrics Metadata
+# Add 1st class feature to PromQL for handling info type metrics
 
 * **Owners:**
   * Arve Knudsen [@aknuds1](https://github.com/aknuds1) [arve.knudsen@grafana.com](mailto:arve.knudsen@grafana.com)
@@ -11,39 +11,61 @@
 * **Other docs or links:**
   * [Proper support for OTEL resource attributes](https://docs.google.com/document/d/1FgHxOzCQ1Rom-PjHXsgujK8x5Xx3GTiwyG__U3Gd9Tw/edit#heading=h.unv3m5m27vuc)
   * [Special treatment of info metrics in Prometheus](https://docs.google.com/document/d/1ebhGNLs3uhdeprJCullM-ywA9iMRDg_mmnuFAQCloqY/edit#heading=h.2rmzk7oo6tu8)
+  * [Scenarios scratch pad](https://docs.google.com/document/d/1nV6N3pDfvZhmG2658huNbFSkz2rsM6SpkHabp9VVpw0/edit#heading=h.luf3yapzr29e)
 
-> This proposal collects the requirements and implementation proposals for enhancing Prometheus with native support for info metrics metadata.
+> This proposal collects the requirements and implementation proposals for adding a 1st class feature to PromQL for handling info type metrics.
 
 ## Why
 
-Currently Prometheus "forgets" which are the identifying labels of info metrics upon ingestion, even though this information is present in at least the OpenMetrics protobuf exposition format (the OpenMetrics text exposition format unfortunately lacks this capability).
-The fact that Prometheus lacks a notion of which are info metrics' identifying labels leads to certain problems:
+Currently, enriching Prometheus query results with corresponding labels from info metrics is challenging.
+More specifically, it requires writing advanced PromQL to join with the info metric in question.
+Take as an example querying HTTP request rates per K8s cluster and status code, while having to join with the `target_info` metric to obtain the `k8s_cluster_name` label:
 
+```promql
+sum by (k8s_cluster_name, http_status_code) (
+    rate(http_server_request_duration_seconds_count[2m])
+  * on (job, instance) group_left (k8s_cluster_name)
+    target_info
+)
+```
+
+The `target_info` metric is in fact the motivation for this proposal, as it's how Prometheus encodes OpenTelemetry (OTel for short) [resource attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md).
+As a result, it's a very important info metric for those using Prometheus as an OTel backend.
+OTel resource attributes model metadata about the environment producing metrics received by the backend (e.g. Prometheus), and Prometheus persists them as labels of `target_info`.
+Typically, OTel users want to include some of these attributes (as `target_info` labels) in their query results, to correlate them with entities of theirs (e.g. K8s pods).
+
+Based on user demand, it would be preferable if Prometheus were to have better UX for enriching query results with info metrics labels, especially with OTel in mind.
+There are other problems with Prometheus' current method of including info metric labels in queries, beyond just the technical barrier:
 * Explicit knowledge of each info metric's identifying labels must be embedded in join queries for when you wish to enrich queries with data (non-identifying) labels from info metrics.
-* Complex join queries must be written in order to enrich time series with corresponding labels from info metrics.
-  This is particularly problematic in the OpenTelemetry (AKA OTel) context, since users depend on (joining with) the `target_info` info metric in order to add relevant [resource attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md) back to their Prometheus metrics.
+  * A certain pair of OTel resource attributes (`service.name` and `service.instance.id`) are currently assumed to be the identifying pair and mapped to `target_info`'s `job` and `instance` labels respectively, but this may become a dynamic property of the OTel model.
+  * Both attributes are in reality optional, so either of them might be empty (`service.name` is only mandatory for OTel SDK clients).
+  * If both identifying attributes are empty, `target_info` isn't generated (there being no identifying labels to join against).
 * If an info metric's data (non-identifying) labels change (a situation that should become more frequent with OTel in the future, as the model will probably start allowing for non-identifying resource attribute mutations), join queries against the info metric (e.g. `target_info`) will temporarily fail due to resolving the join keys to two different metrics, until the old metric is marked stale (by default after five minutes).
 
-Especially in order to provide the best possible OTel experience, the info metric (`target_info` in the case of OTel) staleness problem needs to be solved, so users won't experience temporarily failing join queries while trying to include OTel resource attributes.
-Also, it would be much better if we could provide a simpler query experience where the user doesn't have to know how to write PromQL joins (a fairly complex matter), in order to include e.g. OTel resource attributes.
-Another possible positive outcome might be dedicated support in the Grafana UI for visualizing the resource attributes of each OTel metric.
+If Prometheus could persist info metrics' identifying labels (e.g. `job` and `instance` for `target_info`), human knowledge of the correct identifying labels may become unnecessary when "joining" with info metrics.
+Information about info metric identifying labels is present in at least the OpenMetrics protobuf exposition format (the OpenMetrics text exposition format unfortunately lacks this capability).
+It can also easily be deduced when ingesting metrics from OTLP (OTel Protocol).
+Intrinsic knowledge of info metrics' identifying labels could also help in solving temporary conflicts between old and new versions of info metrics, when data (non-identifying) labels change.
+Another possible positive outcome might be dedicated support in UIs (e.g. Grafana) for visualizing the resource attributes of OTel metrics.
 
 ### Pitfalls of the current solution
 
 Prometheus currently persists info metrics as if they were normal float samples.
-This means that knowledge of info metrics' identifying labels are lost, and you have to base yourself on convention when querying on them (for example that `target_info` should have `job` and `instance` as identifying labels).
+This means that knowledge of info metrics' identifying labels are lost, and you have to base yourself on convention when querying them (for example that `target_info` should have `job` and `instance` as identifying labels).
+There's also no particular support for enriching query results with info metric labels in PromQL.
 The consequence is that you need relatively expert level PromQL knowledge to include info metric labels in your query results; as OTel grows in popularity, this becomes more and more of a problem as users will want to include certain labels from `target_info` (corresponding to OTel resource attributes).
-Without persisted info metric metadata, one can't build more user friendly abstractions (e.g. a PromQL function) for including OTel resource attributes (or other info metric labels) in query results. Neither can you build dedicated UI for OTel resource attributes (or other info metric labels).
+Without persisted info metric metadata, one can't build more user friendly abstractions (e.g. a PromQL function) for including OTel resource attributes (or other info metric labels) in query results.
+Neither can you build dedicated UI for OTel resource attributes (or other info metric labels).
 
 ## Goals
 
 Goals and use cases for the solution as proposed in [How](#how):
 
-* Persist info metrics with known identifying labels as a new info metric sample type.
-* Store for each info metric sample (of the new type) which are the identifying labels.
-* Store in the TSDB immediately that the previous version of an info metric is stale, when its data labels change.
+* Persist info metrics with labels categorized as either identifying or non-identifying.
+* Track when info metrics' set of identifying labels changes. This shouldn't be a frequent occurrence, but it should be handled.
+* Automatically treat the old version of an info metric as stale for query result enriching purposes, when its data labels change (producing a new time series, but with same identity).
 * Add TSDB API for, given a certain time series and a certain timestamp, getting data labels, potentially filtered by certain matchers, from info metrics with identifying labels in common with the time series in question.
-* Simplify inclusion of info metric labels in PromQL.
+* Simplify enriching of query results with info metric labels in PromQL, e.g. via a new function.
 
 ### Audience
 
@@ -53,12 +75,24 @@ Prometheus maintainers.
 
 ## How
 
-* A new info metric sample type will be introduced, where the sample value is the info metric's identifying labels.
-* The head and block indexes will be augmented with indexes of info metrics.
-* A method will be added to the TSDB API for matching info metric data labels to a time series, given a certain timestamp and potentially data label matchers - the method will use the aforementioned head and block info metric indexes.
-* Thanks to the head and block info metric indexes, the info metric staleness problem should be solved, since one can pick the latest version of the info metric for overlapping time ranges.
-* We propose simplifying the inclusion of info metric labels in PromQL through a new `info` function (TODO: describe).
+* Introduce a new info metric sample type, to track the info metric's identifying label set over time (in case it changes).
+* Augment the head and block indexes with indexes of info metrics, for easy finding of info metrics matching time series.
+* Add a method to the TSDB API for matching info metric data labels to a time series, given a certain timestamp and potentially data label matchers - the method will use the aforementioned head and block info metric indexes.
+* Simplify the inclusion of info metric labels in PromQL through a new `info` function: `info(v instant-vector[, ls label-selector])`.
+  This function will be UI for the aforementioned TSDB API.
 
+Using the `info` function, we can simplify the previously given PromQL join example as follows:
+
+```
+sum by (k8s_cluster_name, http_status_code) (
+  info(
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name=~".+"}
+  )
+)
+```
+
+TODO:
 * Make it concise and **simple**; put diagrams; be concrete, avoid using “really”, “amazing” and “great” (:
 * How you will test and verify?
 * How you will migrate users, without downtime. How we solve incompatibilities?
--- a/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
+++ b/proposals/2024-04-10-native-support-for-info-metrics-metadata.md
@@ -38,8 +38,8 @@ Based on user demand, it would be preferable if Prometheus were to have better U
 There are other problems with Prometheus' current method of including info metric labels in queries, beyond just the technical barrier:
 * Explicit knowledge of each info metric's identifying labels must be embedded in join queries for when you wish to enrich queries with data (non-identifying) labels from info metrics.
   * A certain pair of OTel resource attributes (`service.name` and `service.instance.id`) are currently assumed to be the identifying pair and mapped to `target_info`'s `job` and `instance` labels respectively, but this may become a dynamic property of the OTel model.
-  * Both attributes are in reality optional, so either of them might be empty (`service.name` is only mandatory for OTel SDK clients).
-  * If both identifying attributes are empty, `target_info` isn't generated (there being no identifying labels to join against).
+  * Both attributes are in reality optional, so either of them might be missing (`service.name` is only mandatory for OTel SDK clients).
+  * If both identifying attributes are missing, `target_info` isn't generated (there being no identifying labels to join against).
 * If an info metric's data (non-identifying) labels change (a situation that should become more frequent with OTel in the future, as the model will probably start allowing for non-identifying resource attribute mutations), join queries against the info metric (e.g. `target_info`) will temporarily fail due to resolving the join keys to two different metrics, until the old metric is marked stale (by default after five minutes).
 
 If Prometheus could persist info metrics' identifying labels (e.g. `job` and `instance` for `target_info`), human knowledge of the correct identifying labels may become unnecessary when "joining" with info metrics.
@@ -61,9 +61,11 @@ Neither can you build dedicated UI for OTel resource attributes (or other info m
 
 Goals and use cases for the solution as proposed in [How](#how):
 
-* Persist info metrics with labels categorized as either identifying or non-identifying.
+* Persist info metrics with labels categorized as either identifying or non-identifying (i.e. data labels).
 * Track when info metrics' set of identifying labels changes. This shouldn't be a frequent occurrence, but it should be handled.
-* Automatically treat the old version of an info metric as stale for query result enriching purposes, when its data labels change (producing a new time series, but with same identity).
+  * When enriching a query result's labels with data labels from info metrics, it should be considered per timestamp what are each potentially matching info metric's identifying labels (since the identifying label set may change over time).
+* Automatically treat the old version of an info metric as stale for query result enriching purposes, when its data labels change (producing a new time series, but with same identity from an info metric perspective).
+  * When enriching a query result's labels with data labels from info metrics, and there are several matches with equally named info metrics (e.g. `target_info`) for a timestamp, the one with the newest sample wins (others are considered stale).
 * Add TSDB API for, given a certain time series and a certain timestamp, getting data labels, potentially filtered by certain matchers, from info metrics with identifying labels in common with the time series in question.
   * If no data label matchers are provided, _all_ the data labels of found info metrics are added to the resulting time series.
   * If data label matchers are provided, only info metrics with matching data labels are considered.
@@ -74,11 +76,11 @@ Goals and use cases for the solution as proposed in [How](#how):
   * A data label matcher like `__name__="target_info"` can be used to restrict the info metrics used.
     However, the `__name__` label itself will not be copied.
   * Label collisions: The input instant vector could already contain labels that are also part of the data labels of a matching info metric.
-    Furthermore, since the info function might find multiple differently named info metrics with matching identifying labels, those might have overlapping data labels.
-    In this case, the info function has to check if the values of the affected labels match or are different.
+    Furthermore, since multiple differently named info metrics with matching identifying labels might be found, those might have overlapping data labels.
+    In this case, the implementation has to check if the values of the affected labels match or are different.
     The former case is not really a label collision and therefore causes no problem.
-    In the latter case, however, the function has to return an error.
-    The collision can be resolved by constraining the labels via the optional label-selector argument of the info function.
+    In the latter case, however, an error has to be returned to the user.
+    The collision can be resolved by constraining the labels via data label matchers.
     And of course, the user always has the option to go back to the original join syntax (or, even better, avoiding ingesting conflicting info metrics in the first place).
 * Simplify enriching of query results with info metric data (non-identifying) labels in PromQL, e.g. via a new function, based on aforementioned TSDB API.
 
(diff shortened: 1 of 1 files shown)


proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,112 @@
...
+
+Using the `info` function, we can simplify the previously given PromQL join example as follows:
+
+```
+sum by (k8s_cluster_name, http_status_code) (
+  info(
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name=~".+"}
+  )
+)
```
fstab, line 92: Can you help me understand this example? I know the current rule for aggregating rates is [Rate then sum, never sum then rate](https://www.robustperception.io/rate-then-sum-never-sum-then-rate).

Now, will this turn into `rate then info then sum`, or could you also `info then rate then sum`?

Not sure if there is any specific reason to do `rate()` before `info()`, but `sum()` after `info()`.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,112 @@
...
+* **Other docs or links:**
+  * [Proper support for OTEL resource attributes](https://docs.google.com/document/d/1FgHxOzCQ1Rom-PjHXsgujK8x5Xx3GTiwyG__U3Gd9Tw/edit#heading=h.unv3m5m27vuc)
+  * [Special treatment of info metrics in Prometheus](https://docs.google.com/document/d/1ebhGNLs3uhdeprJCullM-ywA9iMRDg_mmnuFAQCloqY/edit#heading=h.2rmzk7oo6tu8)
+  * [Scenarios scratch pad](https://docs.google.com/document/d/1nV6N3pDfvZhmG2658huNbFSkz2rsM6SpkHabp9VVpw0/edit#heading=h.luf3yapzr29e)
+
+> This proposal collects the requirements and implementation proposals for adding a 1st class feature to PromQL for handling info type metrics.
+
+## Why
+
+Currently, enriching Prometheus query results with corresponding labels from info metrics is challenging.
```
bwplotka, line 20: Can we introduce info metric first? Or is it just a gauge now?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Disclaimer: I just copied this example from Beorn's [original document](https://docs.google.com/document/d/1ebhGNLs3uhdeprJCullM-ywA9iMRDg_mmnuFAQCloqY/edit?pli=1#heading=h.37hgysjhge82). I'll answer based on my knowledge of `info` though.

`rate(http_server_request_duration_seconds_count[2m])` becomes a sub-query to the `info` function, which adds the `k8s_cluster_name` label from `target_info` to the `http_server_request_duration_seconds_count` time series resulting from the sub-query. The `sum` expression aggregates by the labels `k8s_cluster_name` and `http_status_code`. If you don't apply `sum` after `info`, the `k8s_cluster_name` label won't be available.

As for whether you could do `info` before `rate`; I haven't tested, but I think it would work. However, it looks less logical/comprehensible to me, as you really want the rate of `http_server_request_duration_seconds_count[2m]`.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,112 @@
...
+### Audience
+
+Prometheus maintainers.
+
+## Non-Goals
+
+## How
+
+* Introduce a new info metric sample type, to track the info metric's identifying label set over time (in case it changes).
+* Augment the head and block indexes with indexes of info metrics, for easy finding of info metrics matching time series.
```
bwplotka, line 79: Would be nice to know how exactly 🙈

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,112 @@
+# Add 1st class feature to PromQL for handling info type metrics
```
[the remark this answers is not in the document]
aknuds1: I don't think "target info" metrics would be fitting, since it's about info metrics in general, not just `target_info`. Also, the current title is geared towards the `info` PromQL function, rather than how to implement support for it (native info metric samples are a way to implement support).

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Yes, I'm trying to derive how to best model the info metric indexes in the [scenarios scratch pad](https://docs.google.com/document/d/1nV6N3pDfvZhmG2658huNbFSkz2rsM6SpkHabp9VVpw0/edit#heading=h.luf3yapzr29e).

I think I need some feedback on whether my idea of tracking changes to info metrics' identifying labels through a native sample type is a good idea. @beorn7 [suggests](https://docs.google.com/document/d/1ebhGNLs3uhdeprJCullM-ywA9iMRDg_mmnuFAQCloqY/edit?pli=1#heading=h.lhskfxuy91g2) that we _might_ persist these info metrics completely outside of the TSDB, but I have the impression that would require reinventing the wheel wrt. querying the info metrics themselves (rather than just including their labels via the `info` function).

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,112 @@
...
+Prometheus currently persists info metrics as if they were normal float samples.
+This means that knowledge of info metrics' identifying labels are lost, and you have to base yourself on convention when querying them (for example that `target_info` should have `job` and `instance` as identifying labels).
+There's also no particular support for enriching query results with info metric labels in PromQL.
+The consequence is that you need relatively expert level PromQL knowledge to include info metric labels in your query results; as OTel grows in popularity, this becomes more and more of a problem as users will want to include certain labels from `target_info` (corresponding to OTel resource attributes).
+Without persisted info metric metadata, one can't build more user friendly abstractions (e.g. a PromQL function) for including OTel resource attributes (or other info metric labels) in query results.
+Neither can you build dedicated UI for OTel resource attributes (or other info metric labels).
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
```
bwplotka, line 62: Is the goal to also select certain metrics based on certain target_info non-identifying label?

e.g. would I be able to do something like `info(http_request{...}, {"go_version" = "1.22.3"}`, so filtering http_requests by label which only exists on target e.g. version of the Go that target was compiled with?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
bwplotka: Exactly yes!

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,112 @@
...
+    rate(http_server_request_duration_seconds_count[2m])
+  * on (job, instance) group_left (k8s_cluster_name)
+    target_info
+)
+```
+
+The `target_info` metric is in fact the motivation for this proposal, as it's how Prometheus encodes OpenTelemetry (OTel for short) [resource attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md).
+As a result, it's a very important info metric for those using Prometheus as an OTel backend.
+OTel resource attributes model metadata about the environment producing metrics received by the backend (e.g. Prometheus), and Prometheus persists them as labels of `target_info`.
+Typically, OTel users want to include some of these attributes (as `target_info` labels) in their query results, to correlate them with entities of theirs (e.g. K8s pods).
```
bwplotka, line 35: What "correlating them" actually means here? Can add some specific examples? Does it only mean joining the data so series have certain extra labels? Or does it mean to also be able to do some further filtering based on that new labels from target_info (e.g. to filter based on target info label?)

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: It would certainly be possible to introduce the necessary info metric modeling before the `info` function, but I think first of all we have to agree on the model :) The `info` function's semantics helps us understand what kind of functionality the model must provide.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: I added more specification about the new TSDB API and its use of label matchers. If I understand you correctly, the answer is yes:

> If data label matchers are provided, time series are only included in the result if matching data labels from info metrics were found.

bwplotka: Nice! Then I wonder if simpler solution would be worth to discuss as an alternative e.g. special prefixed labels: https://cloud.google.com/stackdriver/docs/managed-prometheus/promql#metadata-labels 🤔

aknuds1: I will have a read, thanks.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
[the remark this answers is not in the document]
aknuds1: @beorn7 WDYT?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: > @beorn7 [suggests](https://docs.google.com/document/d/1ebhGNLs3uhdeprJCullM-ywA9iMRDg_mmnuFAQCloqY/edit?pli=1#heading=h.lhskfxuy91g2) that we might persist these info metrics completely outside of the TSDB

Scratch that, in retrospect I noticed a footnote in the doc, where @beorn7 does say we need to keep storing info metric samples in TSDB to support querying:

> To query and display an info metric as a regular time series, we would still store the info metric as usual in the TSDB.

@beorn7 my question is: Does it make sense to you to use the identifying label set as the info metric sample value? I.e., it would be a new info metric sample type, that also pretends to have a float value of 1 (backwards compatible behaviour). From my prototyping, this is what would make sense, otherwise you would need to model changes over time in the info metric index. It seems better to me to model this via the usual TSDB time series mechanism instead.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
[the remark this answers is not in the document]
beorn7: Yeah, this is definitely not restricted to target info.

How about "Simplify joins with info metrics in PromQL"?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
beorn7: I don't have a detailed implementation idea here. You have probably most experience with this by now through your prototyping.

However, it feels like you wanted to say "Does it make sense to you to use the data label set as the info metric sample value?". The identifying labels are like "normal" labels, by which you want to find an info metric, and then you take the data labels from their sample value and add it to the metric you want to join with. Right?

beorn7:
I have just responded to a few direct questions. I haven't reviewed the whole proposal yet. My apologies in case my answers are stupid because I didn't grasp the full context.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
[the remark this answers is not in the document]
aknuds1: Followed your suggestion @beorn7.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: After reading the link about metadata labels, as I understand Bartek, the idea is to perhaps use metadata like syntax instead of introducing an `info` function. I think the underlying TSDB API would be the same. Alternative syntax to `info` is definitely on the table, also @jesusvazquez has thought about this.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,127 @@
...
+Without persisted info metric metadata, one can't build more user friendly abstractions (e.g. a PromQL function) for including OTel resource attributes (or other info metric labels) in query results.
+Neither can you build dedicated UI for OTel resource attributes (or other info metric labels).
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Persist info metrics with labels categorized as either identifying or non-identifying.
+* Track when info metrics' set of identifying labels changes. This shouldn't be a frequent occurrence, but it should be handled.
+* Automatically treat the old version of an info metric as stale for query result enriching purposes, when its data labels change (producing a new time series, but with same identity).
```
codesome, line 66: Let's say a time series starts at time t0, and if the data labels of the info metric changes at time t1, and t2 is a future time t2>t1. Then in this scenario, when I do a range query of info() function for time range t0 through t2, which data labels will be used? Will it use the old data labels between t0 and t1 and then new data labels between t1 to t2, or new data labels for t0 through t2? It will be good to document this because "stale" makes me feel new labels for everything but I think the right thing to do will be old labels for t0 to t1 (I think this is the plan, right? I forgot)

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,127 @@
...
+The consequence is that you need relatively expert level PromQL knowledge to include info metric labels in your query results; as OTel grows in popularity, this becomes more and more of a problem as users will want to include certain labels from `target_info` (corresponding to OTel resource attributes).
+Without persisted info metric metadata, one can't build more user friendly abstractions (e.g. a PromQL function) for including OTel resource attributes (or other info metric labels) in query results.
+Neither can you build dedicated UI for OTel resource attributes (or other info metric labels).
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Persist info metrics with labels categorized as either identifying or non-identifying.
+* Track when info metrics' set of identifying labels changes. This shouldn't be a frequent occurrence, but it should be handled.
```
codesome, line 65: Similar question as the data labels - if the identifying labels change in between let's say t1 (with t0 being a older time and t2 being a future time), do we use the old identifying labels for queries touching t0 to t1, and then new identifying labels for queries after t1? Let's document this.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,127 @@
...
+  * If no data label matchers are provided, _all_ the data labels of found info metrics are added to the resulting time series.
+  * If data label matchers are provided, only info metrics with matching data labels are considered.
+  * If data label matchers are provided, _precisely_ the data labels specified by the label matchers are added to the returned time series.
+  * If data label matchers are provided, time series are only included in the result if matching data labels from info metrics were found.
+  * A data label matcher like `k8s_cluster_name=~".+"` guarantees that each returned time series has a non-empty `k8s_cluster_name` label, implying that time series for which no matching info metrics have a data label named `k8s_cluster_name` (including the case where no matching info metric exists at all) will be excluded from the result.
+  * A special case: If a data label matcher allows empty labels (equivalent to missing labels, e.g. `k8s_cluster_name=~".*"`), it will not exclude time series from the result even if there's no matching info metric.
+  * A data label matcher like `__name__="target_info"` can be used to restrict the info metrics used.
```
codesome, line 74: Do we want to allow writing it as just `target_info`, and with additional matchers like `target_info{k8s_cluster_name=~".+"}`? Or is there a reason why it needs to be `{__name__="target_info"}`?

codesome (commented):
Thanks for writing this up! Overall it's looking good.

I had a discussion with @aknuds1 offline to understand this better and I think the proposal should have these things:
* Details on how info() function will work when you have different info metrics under different names. (the union of data labels that you mentioned)
* Specifics on how a change in identifying and non-identifying labels will impact queries (I left individual comments on them below)
* Any specific details from the linked doc that you think is important, please include them in the proposal because naturally I would consider them as background information and not a part of what you are proposing.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
beorn7: Yes, I think it should be always the relevant label for the respective timestamp. I guess the term "stale" comes from looking at a possible implementation detail, which is to insert a staleness marker for the "old series". However, if the identity of the info series is only defined by the identifying labels, then there is actually no new series. Only the existing series changes its value (where the value are the data labels). (This is the "logical view", if you want. I can imagine it is hard to implement that way directly in Prometheus. And I haven't looked at @aknuds1's PoC to check how it is done there.)

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: The current implementation only recognizes a selector starting with curly braces indeed. But it's also a very rough and hacky implementation, as I'm concentrating on getting the underlying design (TSDB API) right primarily.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: I'm not sure if the prototype currently works this way, since I did some additional thinking when drafting [relevant scenarios](https://docs.google.com/document/d/1nV6N3pDfvZhmG2658huNbFSkz2rsM6SpkHabp9VVpw0/edit#heading=h.h04tn1hwz8ey) for `info`, but my current thinking is indeed that the `info` function should pick the correct data labels on a sample by sample basis. Does that answer your question?

@beorn7 I haven't changed the fact that there will be a new time series when the data labels change. I just use a head/block index to associate info metrics with their identifying label sets, so you can find different time series (as data labels change) for the same identifying labels.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: > do we use the old identifying labels for queries touching t0 to t1, and then new identifying labels for queries after t1?

@codesome yes, that is my current thinking at least. My thinking is to track through the info metric's samples, what is the identifying label set at any given point in time. Agree to document it properly, I'll give my proposal a re-read, to see whether I need to clarify further.

> Maybe a change of identifying labels should really be seen as a completely independent new series, without any staleness marker insertion.

@beorn7 there shouldn't be any need for a staleness marker insertion, when the identifying label set changes, so long as the entire series' label set remains the same. My thinking is to just record a new sample for the info metric, that records the new set of identifying labels. Does that make sense to you as well?

aknuds1:
Thanks for the review @codesome (and for our chat beforehand)!

> Details on how info() function will work when you have different info metrics under different names. (the union of data labels that you mentioned)

I'll look the proposal over again, and write more if this is missing.

> Specifics on how a change in identifying and non-identifying labels will impact queries (I left individual comments on them below)

Thanks, I'll keep this in mind when re-reading the proposal also.

> Any specific details from the linked doc that you think is important, please include them in the proposal because naturally I would consider them as background information and not a part of what you are proposing.

Agree 100%.

aknuds1:
> Specifics on how a change in identifying and non-identifying labels will impact queries (I left individual comments on them below)

@codesome I added more details to the How section:

> * Augment the head and block indexes with indexes of info metrics, for easy finding of info metrics matching time series.
    * The TSDB head and every block register their respective info metrics in a corresponding index, with the different identifying label sets each info metric has had over its lifetime.
> * Add a method to the TSDB API for matching info metric data labels to a time series, given a certain timestamp and potentially data label matchers - the method will use the aforementioned head and block info metric indexes.
    * Candidate info metrics are found by searching the info metric index for info metrics with identifying labels contained in the input label set.
      * Each candidate info metric's identifying label set _for the timestamp in question_, is obtained from the info metric's samples.
      * If that identifying label set is not a match, the info metric is ignored.
      * If several info metrics with the same name are found, the one with the latest sample is chosen (i.e., older metrics are considered stale).
    * Data labels are picked from the found info metrics according to the rules defined in the Goals section.

Does that work for you?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: > How to implement this is again a question that you can probably answer better than anyone by now. It depends how "thoroughly" the new data model will be implemented in PromQL.

I've thus far chosen the "simplistic" route of not changing Prometheus' core time series identity model (so that when the identifying label set changes, as you describe above, it's the same time series according to Prometheus). Then, for the purpose of implementing the `info` function, the info metric's identifying label set is determined on a per timestamp basis (via the info metric's samples). I haven't implemented the latter, it's just design at the moment, but I think it should work.

You don't see any real problems with my proposed design, as I describe it above, I hope?

beorn7: I cannot say at the moment.

In any case, changing which labels are identifying should be a rare thing to happen, so even if it isn't solved in the ideal way, it shouldn't hurt much.

This is very much different from the current state of a change in a target_info label that then creates duplicates because staleness isn't handled properly when ingesting OTLP. The latter happens _all the time_ in regular operation.

aknuds1: Thanks! :) I'll try fixing up the prototype to match the design in the proposal, then.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+  )
+)
+```
+
+TODO:
+
+* Specify detection of info metric identifying labels for other ingestion methods than OTLP.
+* Define how this functionality would work together with OOO samples.
+
+## Alternatives
```
beorn7, line 125: This could describe what I alluded to above, i.e. "adding all resource attributes and build versions and pod labels and everything as prefixed labels" with its pros and cons.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+    However, the `__name__` label itself will not be copied.
+  * Label collisions: The input instant vector could already contain labels that are also part of the data labels of a matching info metric.
+    Furthermore, since multiple differently named info metrics with matching identifying labels might be found, those might have overlapping data labels.
+    In this case, the implementation has to check if the values of the affected labels match or are different.
+    The former case is not really a label collision and therefore causes no problem.
+    In the latter case, however, an error has to be returned to the user.
+    The collision can be resolved by constraining the labels via data label matchers.
+    And of course, the user always has the option to go back to the original join syntax (or, even better, avoiding ingesting conflicting info metrics in the first place).
+* Simplify enriching of query results with info metric data (non-identifying) labels in PromQL, e.g. via a new function, based on aforementioned TSDB API.
+
```
ying-jeanne, line 86: I would like to clarify whether `ensure backward compatibility with current Prometheus usage and minimize potential conflicts with existing metric labels` is part of the goal. especially when considering metadata labels as an alternative solution, this clarity will guide whether this approach is viable and if additional tools or documentation are needed to guide user managing such conflicts.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+Prometheus maintainers.
+
+## Non-Goals
+
+## How
+
+* Introduce a new info metric sample type, to track the info metric's identifying label set over time (in case it changes).
+* Augment the head and block indexes with indexes of info metrics, for easy finding of info metrics matching time series.
+  * The TSDB head and every block register their respective info metrics in a corresponding index, with the different identifying label sets each info metric has had over its lifetime.
+* Augment the OTLP endpoint to specify `target_info`'s identifying labels when ingesting write requests, and to store it as the native info metric type.
```
ying-jeanne, line 98: Question: Is this proposal only focusing on otlp endpoint? otherwise it might worth mentioning all ingestions: openmetrics, otlp, classic prometheus exposition formats and remote write.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Makes sense @beorn7, I'll try to add to the introduction as you suggest.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Thanks for sharing those insights @beorn7 :)

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+    The collision can be resolved by constraining the labels via data label matchers.
+    And of course, the user always has the option to go back to the original join syntax (or, even better, avoiding ingesting conflicting info metrics in the first place).
+* Simplify enriching of query results with info metric data (non-identifying) labels in PromQL, e.g. via a new function, based on aforementioned TSDB API.
+
+### Audience
+
+Prometheus maintainers.
+
+## Non-Goals
+
```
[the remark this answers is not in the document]
aknuds1: Removed it.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+  info(
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name=~".+"}
+  )
+)
+```
+
+TODO:
+
+* Specify detection of info metric identifying labels for other ingestion methods than OTLP.
```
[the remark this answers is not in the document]
aknuds1: Thanks for the kind suggestion :)

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Good suggestion.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Thanks, this is good feedback. It's good to explicitly call out an implicit motivation behind the `info` function design, i.e. backwards compatibility (improving the existing solution of info metrics).

> minimize potential conflicts with existing metric labels

Do you mean here that adding e.g. OTel resource attributes as labels could lead to conflicts?

I've added the following two points to goals:

* Ensure backwards compatibility with current Prometheus usage
* Minimize potential conflicts with existing metric labels

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+* Explicit knowledge of each info metric's identifying labels must be embedded in join queries for when you wish to enrich queries with data (non-identifying) labels from info metrics.
+  * A certain pair of OTel resource attributes (`service.name` and `service.instance.id`) are currently assumed to be the identifying pair and mapped to `target_info`'s `job` and `instance` labels respectively, but this may become a dynamic property of the OTel model.
+  * Both attributes are in reality optional, so either of them might be missing (`service.name` is only mandatory for OTel SDK clients).
+  * If both identifying attributes are missing, `target_info` isn't generated (there being no identifying labels to join against).
+* If an info metric's data (non-identifying) labels change (a situation that should become more frequent with OTel in the future, as the model will probably start allowing for non-identifying resource attribute mutations), join queries against the info metric (e.g. `target_info`) will temporarily fail due to resolving the join keys to two different metrics, until the old metric is marked stale (by default after five minutes).
```
dashpole, line 43: You may be interested in the current OTel Entities proposal, which is designed to carry non-identifying resource information: https://github.com/open-telemetry/oteps/blob/6d6febfaf05f130c703e2dd0fa91dfacada82a7d/text/entities/0256-entities-data-model.md.  My hope would be that we can potentially map these to OM info metrics in the future.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+
+If Prometheus could persist info metrics' identifying labels (e.g. `job` and `instance` for `target_info`), human knowledge of the correct identifying labels may become unnecessary when "joining" with info metrics.
+Information about info metric identifying labels is present in at least the OpenMetrics protobuf exposition format (the OpenMetrics text exposition format unfortunately lacks this capability).
```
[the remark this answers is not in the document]
dashpole: I also worry about this being limited to OM, as most of the ecosystem seems to still be using prometheus text/protobuf formats today.

Did we consider to expanding this to also include gauge metrics with an `_info` suffix?  Many OTel prometheus exporters still export target_info as a gauge due to lack of OM support, or because the Prometheus client doesn't support Info metrics.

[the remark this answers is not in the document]
beorn7: Even OM needs an amendment for its text version.

So the best outcome here is to "fix" OpenMetrics (including marking identifying labels in the text format and allowing info metrics that do not end on _info, but also a number of other things that have hampered adoption so far) so that it finally OM gets adopted everywhere.

The second best outcome is to also amend classic Prometheus exposition formats. (It's a bit weird because I'd rather see the effort of changing the exposition format to lead directly to OM everywhere rather than more versions of the many formats that we already have.)

In the meantime, I would go for scrape config options to mark metrics of whatever source and format as info metrics, including the option to mark certain labels as identifying labels. (The heuristic "target labels → identifying labels, other labels → date labels" should be fine in most cases.) A rule to say "all metrics ending on `_info` are info metrics" would be easy to configure. And then special cases could be configured to say "`kube_pod_labels` coming from kube-state-metrics should be considered an info metric with `namespace` and `pod` as the identifying labels. Etc.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: I've fixed up the prototype, so it matches the design in the proposal (info metric samples define the identifying label set).

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+Typically, OTel users want to include some of these attributes (as `target_info` labels) in their query results, to correlate them with entities of theirs (e.g. K8s pods).
+
+Based on user demand, it would be preferable if Prometheus were to have better UX for enriching query results with info metrics labels, especially with OTel in mind.
+There are other problems with Prometheus' current method of including info metric labels in queries, beyond just the technical barrier:
+* Explicit knowledge of each info metric's identifying labels must be embedded in join queries for when you wish to enrich queries with data (non-identifying) labels from info metrics.
+  * A certain pair of OTel resource attributes (`service.name` and `service.instance.id`) are currently assumed to be the identifying pair and mapped to `target_info`'s `job` and `instance` labels respectively, but this may become a dynamic property of the OTel model.
+  * Both attributes are in reality optional, so either of them might be missing (`service.name` is only mandatory for OTel SDK clients).
```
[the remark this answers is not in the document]
aknuds1: Even after that PR, `service.instance.id` is only _recommended_, right? Is there a plan to make `service.instance.id` required for OTel SDK clients? Also, it doesn't change the situation for non-SDK clients AFAICT.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: @dashpole yes, this point in fact refers to the proposed Entity model, since it will allow for non-identifying resource/entity attributes to change.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Done @beorn7 - PTAL.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,136 @@
...
+  * If both identifying attributes are missing, `target_info` isn't generated (there being no identifying labels to join against).
+* If an info metric's data (non-identifying) labels change (a situation that should become more frequent with OTel in the future, as the model will probably start allowing for non-identifying resource attribute mutations), join queries against the info metric (e.g. `target_info`) will temporarily fail due to resolving the join keys to two different metrics, until the old metric is marked stale (by default after five minutes).
+
+If Prometheus could persist info metrics' identifying labels (e.g. `job` and `instance` for `target_info`), human knowledge of the correct identifying labels may become unnecessary when "joining" with info metrics.
+Information about info metric identifying labels is present in at least the OpenMetrics protobuf exposition format (the OpenMetrics text exposition format unfortunately lacks this capability).
+It can also easily be deduced when ingesting metrics from OTLP (OTel Protocol).
```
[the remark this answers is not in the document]
aknuds1: I'm adding the following sentence, does it work for you @beorn7?

> Most info metrics' identifying labels will be `job` and `instance`, but there are some exceptions (e.g. `kube_pod_labels`).

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: I quickly drafted a corresponding alternative, it could be expanded upon. PTAL.

gouthamve:
I commented on [Proper support for OTEL resource attributes](https://docs.google.com/document/d/1FgHxOzCQ1Rom-PjHXsgujK8x5Xx3GTiwyG__U3Gd9Tw/edit#heading=h.unv3m5m27vuc) but also putting it here as this is being pitched as a solution for OTel's resource attribute handling.

I feel like we're forgetting the equivalent problem is the Pull model. Service Discovery gives us a TON of labels too and these are the resource attribute equivalents, and we don't actually copy them ALL over or put them in an `_info` function. We make the users explicitly pick a set of labels that they'd like copied into the metrics and this model has worked well for us all these years.

All we need to do solve the OTel problem is make the users pick a set of labels (proposal 1. in the doc) and I think we should be good tbh. It brings it in line with how service discovery works today.

But just `info` function without this is not adequate imo.

-----

Coming to the `info` function, it is still useful. Infact it makes it possible to emit a `service_discovery_info` metric that has _ALL_ the labels exposed by the SD component and `info` makes it easier to join on those.

aknuds1:
@gouthamve Would you like to propose that as one of the alternatives in this proposal? I think we could solve that by your opening a PR to my branch, with the proposed alternative.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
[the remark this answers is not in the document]
dashpole: Thats correct.  SDKs will start generating it by default, but there are ways to get around it, and the collector won't always send it.  It will get better, but not be solved entirely

aknuds1:
After discussing with @codesome, I've pushed a revision that leaves TSDB details out of this proposal. This is because @codesome wants for us to consider the `info` design independently of TSDB changes, and instead have a separate proposal for changes to TSDB (info metric indexes, native info metric sample type).

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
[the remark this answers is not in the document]
jsuereth: We need to resolve this in the entity WG somehow (coordinating with prometheus).

Ideally we have a way to give you a grouping (job) and unique identifier (instance) for any piece of telemetry you receive from OTEL.

At a minimum, I think we should be able to provide a unique identifier (instance), where "job/instance" matching should always work.

As a path forward, we'll take this as a constraint on Entity work to make sure we can fulfill this condition.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,154 @@
...
+Another possible positive outcome might be dedicated support in UIs (e.g. Grafana) for visualizing the resource attributes of OTel metrics.
+
+### Pitfalls of the current solution
+
+Prometheus currently persists info metrics as if they were normal float samples.
+This means that knowledge of info metrics' identifying labels are lost, and you have to base yourself on convention when querying them (for example that `target_info` should have `job` and `instance` as identifying labels).
+There's also no particular support for enriching query results with info metric labels in PromQL.
+The consequence is that you need relatively expert level PromQL knowledge to include info metric labels in your query results; as OTel grows in popularity, this becomes more and more of a problem as users will want to include certain labels from `target_info` (corresponding to OTel resource attributes).
+Without persisted info metric metadata, one can't build more user friendly abstractions (e.g. a PromQL function) for including OTel resource attributes (or other info metric labels) in query results.
+Neither can you build dedicated UI for OTel resource attributes (or other info metric labels).
```
jesusvazquez, line 64: Perhaps you could mention that there is room for a future iteration addressing these issues in various forms, one could think of info metrics with persistence or some other kind of metadata store.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Persist info metrics with labels categorized as either identifying or non-identifying (i.e. data labels).
```
jesusvazquez, line 70: I think this statements sounds a bit contradictory to the above:

> This means that knowledge of info metrics' identifying labels are lost, ...

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,154 @@
...
+  )
+)
+```
+
+TODO:
+
+* Specify detection of info metric identifying labels for other ingestion methods than OTLP.
+* Define how this functionality would work together with OOO samples.
+
+## Alternatives
```
fpetkovski, line 132: This might have been discussed before, but I wonder if we are missing the simplest solution of query rewriting as a considered alternative. Given the example query on line [119](https://github.com/prometheus/proposals/pull/37/files#diff-e74174dfba7ed37042ee0c8fef56c9fad81cab4fc448067c3477488b1136fba0R119), can we rewrite it into a join and let the engine execute it with the existing semantics of PromQL. Information needed from the join can either be defined implicitly or specified by the user as arguments.

fpetkovski (commented):
My main concern with the proposal is that it starts by describing a problem which can be solved by adding syntactic sugar for a specific type of joins. It then goes into various details on what changes are needed to integrate with OTel and various modifications in the the TSDB. 

If the goal is to enable better support for OTEL `target_info` metrics, then I think this should be emphasized in the title and very clear from the description. Unless OTEL is the primary driver of the feature, I am not convinced that the new `info` syntax/function is better than the explicit join.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: The `info` function can't be simply reduced to syntactic sugar for joins though, for the following reasons at least:

1. The `info` function automatically finds suitable info metrics to join with (the MVP only supports `target_info`, but we aim to lift that limitation in the future)
1. The `info` function automatically knows which are info metrics' identifying labels, i.e. which to join on (the MVP assumes `instance` and `job`, but we aim to lift that limitation in the future)
1. The `info` function solves the problem of temporary conflicts between info metrics due to non-identifying labels changing, until the old version goes stale

> Information needed for the join can either be implicitly deducted or specified by the user as arguments.

Assuming we were to use query rewriting, how would you deduce which info metric to join against (when `info` is no longer limited to just the `target_info` metric)? If the user, as in join queries, has to supply which info metric to join with and which labels to join on plus which labels to include, have we really gained much versus status quo?

aknuds1:
> My main concern with the proposal is that it starts by describing a problem which can be solved by adding syntactic sugar for specific type of joins.

@fpetkovski in my [reply](https://github.com/prometheus/proposals/pull/37#discussion_r1687972402) to your comment, I describe how `info` can't just be reduced to syntactic sugar.

> If the goal is to enable better support for OTEL target_info metrics, then I think this should be emphasized in the title and very clear from the description.

It's definitely a motivation for developing the `info` function, but the overarching goal is to simplify inclusion of data labels from info metrics in general.

@beorn7 would you like to share your PoV as far as the motivation goes?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: @jesusvazquez should that perhaps be mentioned in the Goals section? Before revising the proposal after @codesome's feedback, solving these problems in Prometheus was part of the goals. My understanding of the "Pitfalls of the current solution" section is that it should be referring to status quo in Prometheus, before the proposed solution.

To be honest I'm a bit unsure of what to put in the Pitfalls section currently, until @codesome reviews so the final form of the proposal is clearer.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: I suspect you have a different understanding of the "Pitfalls of the current solution" section than I do. That section has the following purpose (from the template):

> What specific problems are we hitting with the current solution? Why it’s not enough?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,154 @@
...
+* Temporary series churn when migrating existing metrics to this new scheme
+* Increased series churn when metadata labels change
+* More labels per metric increases CPU/memory usage
+
+## Action Plan
+
+The tasks to do in order to migrate to the new idea.
+
+* [ ] Task one <GH issue>
+* [ ] Task two <GH issue> ...
```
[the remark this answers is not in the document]
aknuds1: Thanks, I'll add some.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
MichaHoffmann: I can only see this working well for a static list of well known info metrics like.. `target_info` or by specifying which info metrics to use expicitely. If i have multiple info metrics that match by accident and dont specify matchers I probably get confusing labels added. Thats not possible if i specify the info metric explicitely or constrain it to a well known list like `target_info`. Prometheus is too flexible to make such convenience safe I think.

aknuds1: > I can only see this working well for a static list of well known info metrics like.. target_info or by specifying which info metrics to use expicitely.

@MichaHoffmann if that is your concern with the `info` function (and @fpetkovski's?), I think that should be an explicit criticism of the proposed design, rather than suggesting `info` as proposed here can be simply reduced to syntactic sugar. If you think the design should be explicit about picking info metrics rather than automatic (although that can be filtered down via the `__name__` label matcher), I think there should be an alternative proposal. I'm not sure though if it's most feasible to list competing proposals in the same document, or whether they should be separate.

MichaHoffmann: Oh, Sorry I didnt want to propose something differently, I was just concerning out loud. What I want to say is that the name label matcher for info metric probably should be requirement for this to be generally least surprising for the user.

aknuds1: Thanks, I can note your wish about the name label matcher being a requirement, so a consensus can be reached.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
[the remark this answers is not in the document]
aknuds1: I added two tasks, could you take a look?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
beorn7: I should note that my original idea was indeed that there is generally a relatively low number of info metrics so that the default matching behavior is "all the info metrics known to the TSDB", which is in fact one of the reasons that makes the "joining" easier.

Note that with a "general metadata store", you would probably also end up with "give me all the metadata you know for this metric", which is (more or less) equivalent to "give me all the info metrics that have an identifying labels match with this metric".

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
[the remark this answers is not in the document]
aknuds1: @jsuereth the `info` MVP, as it stands, does assume that identifying labels are `job` and `instance`, but in the future it should be able to detect which are an info metric's identifying labels. The MVP avoids trying to detect the identifying labels, since it'll require persisting knowledge of them (which is a bigger problem to solve).

As I see it currently at least, the main thing with the Entity model in this respect should be that OTLP write requests always specify identifying Resource attributes in a sensible way, so that Prometheus can map them to corresponding labels. WDYT?

> At a minimum, I think we should be able to provide a unique identifier (instance), where "job/instance" matching should always work.

I think at the minimum providing a unique `instance` (or equivalent) identifying resource attribute should work well from Prometheus' PoV at least.

fpetkovski:
Thanks for the explanation @beorn7 and @aknuds1. This does indeed clarify the motivation of the proposal.

aknuds1:
@fpetkovski great :)

aknuds1:
@MichaHoffmann I added [an alternative](https://github.com/prometheus/proposals/pull/37/files#diff-e74174dfba7ed37042ee0c8fef56c9fad81cab4fc448067c3477488b1136fba0R149) to represent your suggestion of making `info` require the user to specify the info metric(s) to join with.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,166 @@
...
+
+* Persist info metrics with labels categorized as either identifying or non-identifying (i.e. data labels).
+* Track when info metrics' set of identifying labels changes. This shouldn't be a frequent occurrence, but it should be handled.
+  * When enriching a query result's labels with data labels from info metrics, it should be considered per timestamp what are each potentially matching info metric's identifying labels (since the identifying label set may change over time).
+* Automatically treat the old version of an info metric as stale for query result enriching purposes, when its data labels change (producing a new time series, but with same identity from an info metric perspective).
+  * When enriching a query result's labels with data labels from info metrics, and there are several matches with equally named info metrics (e.g. `target_info`) for a timestamp, the one with the newest sample wins (others are considered stale).
+* Add TSDB API for, given a certain time series and a certain timestamp, getting data labels, potentially filtered by certain matchers, from info metrics with identifying labels in common with the time series in question.
```
codesome, line 75: I would avoid proposing a TSDB API here to have enough freedom later to make changes later, and instead focus on explaining the working of the info() function, which the points under this TSDB API section actually end up explaining. So I would move all the explanation of the matchers under the "How" section after you introduce the info() function, which will be like explaining the semantics of the info() function. TSDB implementation will naturally tailor to info() function requirements and don't need to define the TSDB changes here.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,166 @@
...
+
+Prometheus maintainers.
+
+## How
+
+* Track the info metric's identifying label set over time (in case it changes) - storage model details to be elaborated in separate proposal.
+* Keep info metric indexes in storage - storage model details to be elaborated in separate proposal.
+  * Info metric indexes maintaining per info metric the different identifying label sets it has had over its lifetime.
+  * Indexing the different identifying label sets an info metric has had over its lifetime allows determining which are potential matches for a given metric, before considering the time dimension.
+* Add a method to the TSDB API for matching info metric data labels to a time series, given a certain timestamp and potentially data label matchers - the method will use the aforementioned info metric indexes.
```
codesome, line 105: Similar comment about TSDB API. Let's extract these points from here and turn it into an explanation of working of info() function. TSDB API will take it's shape similarly and don't need to define it here.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,166 @@
...
+```
+sum by (k8s_cluster_name, http_status_code) (
+  info(
+    rate(http_server_request_duration_seconds_count[2m]),
+    {k8s_cluster_name=~".+"}
+  )
+)
+```
+
+TODO:
```
codesome, line 127: Are these still a TODO? 

I believe detecting identifying labels is something to be taken care on OTel and OpenMetrics side. So I would remove that as a TODO here.

And we can probably remove the OOO point here since that might be more of an implementation detail (if not, we can revisit it later).

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,166 @@
...
+* Simplicity, removes need for joining with info metrics
+
+#### Cons
+
+* Metrics will have potentially far more labels than what's strictly necessary to identify them
+* Temporary series churn when migrating existing metrics to this new scheme
+* Increased series churn when metadata labels change
+* More labels per metric increases CPU/memory usage
+
+### Make the `info` function require specifying the info metric(s)
```
codesome, line 149: I was thinking having an option to specify info metrics to join is not a bad idea. We can still by default join with all info metrics, but have an option to specify a subset as well. Maybe that is something we can add later in a backward compatible fashion. Any thoughts on this @aknuds1 @beorn7?

codesome (commented):
Looking good! My major comments are around leaving out the TSDB API and turning those points into explaining the semantics of info() function instead (which the TSDB API explanation are indirectly doing).

I also have a thought/question on able to specify info metrics to join with, with default being join with all.

Other than that, it is 👍

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
beorn7: Yeah, me thought was that the `info` function should focus on the cases where it "magically" works, and if things get complicated (with label collisions, too many info metrics that match, …), then you can just revert to the traditional join syntax, which is mostly so complex because it specifies everything explicitly. Or in other words: We are diluting the added value of the info function if we add some of the knobs that the traditional join syntax has (and which makes it so unwieldy).

But we can still have this alternative here for public consideration (and to show that this thought was considered).

MichaHoffmann: I guess it would be reasonable if this only works for info metrics that were present in the same scrape maybe?

beorn7: I don't think we should have that restriction. `kube_pod_labels` is an example where you might apply things cross-scrape. OTLP ingestion is an example where there isn't even a notion of "scrape".

For metrics ingested via a scrape, the target labels will be identifying labels by default anyway, so the whole proposed semantics of the info function will implicitly limit info metric matching to the same scrape in most cases (but that's a behavior that needs to be configurable in a different way, e.g. if scraping KSM, one might want to not have the KSM target labels as identifying labels to enable `kube_pod_labels` matching cross-target.)

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: > I believe detecting identifying labels is something to be taken care on OTel and OpenMetrics side. So I would remove that as a TODO here.

I guess that makes sense.

> And we can probably remove the OOO point here since that might be more of an implementation detail (if not, we can revisit it later).

I guess that is indeed an implementation detail, I can remove it.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: > I was thinking having an option to specify info metrics to join is not a bad idea.

@codesome @beorn7 the proposal suggests the ability to specify certain info metrics by matching on `__name__`:

>  * A data label matcher like `__name__="target_info"` can be used to restrict the info metrics used.

Did you consider this?

Maybe there could also be a configuration parameter to control this behaviour?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: Thanks for the feedback, I moved the matchers explanation into the How section, as sub-points to the `info` function. PTAL.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
aknuds1: I moved these points after the `info` function point in the How section, except for the TSDB API point, which I dropped. PTAL.

codesome (approved):
LGTM from me!

@bwplotka @beorn7 @jesusvazquez do you have any feedback? I think we can move forward with the current state since this will start off as an experimental feature, allowing us to make any modifications if necessary (especially with ongoing metadata work on the side).

jesusvazquez (approved):
LGTM! I'm good too lets get this merged 🚀

bwplotka:
Cool, I would love to check it with @dashpole this week.

My main concern is that I have several ppl me DM-ing "have you seen info function proposal? is this amount of complexity only to eliminate one join? is this Otel driven only?". Ideally those would lay out their concerns in this proposal, and will encourage them, but I would love to resolve those concerns before we finalize things, so we don't have complains or repeated discussions like with UTF-8

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,153 @@
...
+Information about info metric identifying labels is present in at least the OpenMetrics protobuf exposition format (the OpenMetrics text exposition format unfortunately lacks this capability).
+It can also easily be deduced when ingesting metrics from OTLP (OTel Protocol).
+Most info metrics' identifying labels will be `job` and `instance`, but there are some exceptions (e.g. `kube_pod_labels`).
+Intrinsic knowledge of info metrics' identifying labels could also help in solving temporary conflicts between old and new versions of info metrics, when data (non-identifying) labels change.
+Another possible positive outcome might be dedicated support in UIs (e.g. Grafana) for visualizing the resource attributes of OTel metrics.
+
+### Pitfalls of the current solution
+
+Prometheus currently persists info metrics as if they were normal float samples.
+This means that knowledge of info metrics' identifying labels are lost, and you have to base yourself on convention when querying them (for example that `target_info` should have `job` and `instance` as identifying labels).
```
bwplotka, line 60: Well, the damage is much bigger -- we essentially use all labels as "identifying" causing massive churn and cardinality bomb. With this feature we essentially increase use of info metric, so that's perhaps relevant to fix.

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
```diff
@@ -0,0 +1,153 @@
...
+There's also no particular support for enriching query results with info metric labels in PromQL.
+The consequence is that you need relatively expert level PromQL knowledge to include info metric labels in your query results; as OTel grows in popularity, this becomes more and more of a problem as users will want to include certain labels from `target_info` (corresponding to OTel resource attributes).
```
bwplotka, line 62: > you need relatively expert level PromQL knowledge to include info metric labels in your query results

From the WHY section I read only one problem. Not knowing what are identifying labels, right? Is that the only problem causing the need of "relatively expert level PromQL knowledge"?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
bwplotka: Thanks!

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
beorn7: Currently, you have to explicitly pick all three of:
1. What labels to join on (AKA identifying labels).
2. What metric to join with.
3. What labels to copy over.

The basic idea is to provide a convenient way for all of this, that works most of the time:
1. Prometheus knows the identifying labels.
2. Join with all matching info metrics (but this can optionally be restricted)
3. Copy all data labels over (but this can optionally be restricted)

While we are on it, we'll also make filtering simpler.

And as a byproduct, we solve the staleness problem we see with current target_info metrics ingested via OTLP. (But to solve just the latter problem, we could probably craft something simpler.)

beorn7:
Thanks for all the discussion here.

And there is even more discussion happening in other channels.

I'm at the same time intrigued and scared by all this growing attention for this (which just started with a small idea of mine). :)

Let's clarify that this is all still an experiment. We want to implement it (in prototypical form) behind a feature flag to give users the option to try it out and compare it to alternatives. We are still open to changing details on this, or even scrap the whole idea altogether. We pretty much agreed elsewhere, that we should also implement mapping of OTel resource attributes to labels upon OTLP ingestion as a parallel experiment.

With that said, I think we should merge this PR in the current form and not fine-tune it based on the outcome of the still ongoing discussions. I think those discussion have reached a point where we really would like to see the results from the prototype in the wild to come to further conclusions.

@jesusvazquez and @codesome have already approved this. @bwplotka as you added some significant comments to this, are you fine with merging or would you prefer more changes on the design doc before merging?

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
bwplotka: Makes sense, thanks!

bwplotka (approved):
Thanks for waiting a bit!

We just had a review cycle with my colleagues about this. Generally, great work on making it fit info metric in existing gauge style and flexibility of Prometheus! Key things worth to check from our side when we are looking on info function as in example:

```
sum by (k8s_cluster_name, http_status_code) (
  info(
    rate(http_server_request_duration_seconds_count[2m]), # original metric to join into
    {k8s_cluster_name=~".+"} # info metric matchers
  )
)
```

1) When user wants to specify one metric, they have to use `__name__` syntax or can they directly do normal PromQL metric matching as is

```
sum by (k8s_cluster_name, http_status_code) (
  info(
    rate(http_server_request_duration_seconds_count[2m]), # original metric to join into
    target_info{k8s_cluster_name=~".+"} 
  )
)
```
?

2) I assume for [MVP](https://github.com/prometheus/proposals/pull/37#discussion_r1687972402) the only metric matched in the second argument is a hardcoded `target_info`, but long term any `{__name__=~"magic.*"` will be supported, right?
  a)  Am I right that this would mean essentially `{__name__=~".*_info"` by default if name is not specified?
  b)  If the MVP is the case why not starting with the mandatory syntax of having metric name specified from the start (not supporting more than one metric for now?)
  
...generally not every system can support flexible metric name matching, but Prometheus can, so it's fair to do this. We will have to probably restrict this a bit.

3) Naming is hard 🙈  `info` name reads like "give me information about this metric", but actually it is "join this labels from another metric".  Should we call it `info_join` or `more_info` or something more explicit? (: 

Thanks for considering, otherwise LGTM 💪🏽

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
beorn7: Maybe we can, at the end of the day, use a normal vector matcher after all, just that the PromQL engine has to intervene before it is fully evaluated.

Yesterday, I discussed with @juliusv his valid concern about using a vector matcher here in a non-standard way, and we found that you could stretch things just a tiny bit to see this vector matcher as a normal vector matcher. It will (potentially) select a lot, but all the info metrics we want will be selected, too, so one might argue the `info` function simply trims this selection down to the relevant info metrics. Of course, for efficiency reasons, we want to do this trimming proactively and not first load all the metrics matching the vector matcher into memory, but that's an optimization that could also be applied to any other PromQL expression without breaking any semantics. The `info` function also looks back at the originally specified labels to find out what labels to copy. But that doesn't break the semantics, either.

juliusv: > The info function also looks back at the originally specified labels to find out what labels to copy.

Yeah, the absent function already does similar magic.

Now on a language typing level, we'll still have the problem that we only have four types: strings, scalars, and instant/range vectors. We currently don't have any way to specify that a function expects not just one of those four types as an input, but a specific AST node type (specifically an instant vector selector, not just any instant-vector-typed result). I guess for a start we could ignore this in parsing and have the function check at run-time whether the parameter is really an instant vector selector and error out otherwise.

aknuds1:
@bwplotka thanks very much for your input/consideration, attempting to answer below:

> 1) When user wants to specify one metric, they have to use __name__ syntax or can they directly do normal PromQL metric matching as is 
> ```
> sum by (k8s_cluster_name, http_status_code) (
>  info(
>    rate(http_server_request_duration_seconds_count[2m]), # original metric to join into
>    target_info{k8s_cluster_name=~".+"}
>  )
> )
> ```

Aha good point. I think this question has been raised by @beorn7 before, but it's not even clear exactly what the data label selector syntax will end up being, since the `{k8s_cluster_name=~".+"}` syntax is recognized by the parser as `VectorSelector`, and we _might_ need a different syntax in order to have a more specialized PromQL expression suitable to `info`. Please take the current data label selector syntax as an idea, that might not work in practice. I think _one_ mentioned possibility would be to provide a variable number of data label matchers directly (without enclosing curly braces). I might experiment with this change and see if I can produce a working parser.

> 2) I assume for [MVP](https://github.com/prometheus/proposals/pull/37#discussion_r1687972402) the only metric matched in the second argument is a hardcoded `target_info`, but long term any `{__name__=~"magic.*"` will be supported, right?
  a)  Am I right that this would mean essentially `{__name__=~".*_info"` by default if name is not specified? Is there a room to keep the possible metric names set somewhere.. to few?
  b)  If the MVP is the case why not starting with the mandatory syntax of having metric name specified from the start (not supporting more than one metric for now?)

My current thinking is that either `info` has to be hardcoded to only respect `target_info` (as it has known identifying labels: `instance`/`job`), or be configurable wrt. info metric names and identifying labels.

Regarding a) and b), I think that specifying the info metric name in PromQL should not be supported in the MVP. I've made a [PoC](https://github.com/prometheus/prometheus/pull/14634) for automatic info metric data label inclusion, that relies on configuration to teach it about info metrics and their respective identifying labels. The benefit of this scheme is that you don't need to make Prometheus intelligent about these things.

> 3) Naming is hard 🙈  `info` name reads like "give me information about this metric", but actually it is "join this labels from another metric".  Should we call it `info_join` or `more_info` or something more explicit? (: 

Don't know, but I'm sure we can discuss this during the development process. We're not even 100% sure it will be a function syntactically (maybe we can think of some more optimal syntax in the end).

bwplotka:
Got it, thanks! Let's iterate then, especially for known number (or only one for now) of info metrics, LGTM!

Thanks for extensive work on this!

proposals/2024-04-10-native-support-for-info-metrics-metadata.md
beorn7: `absent()` is indeed a good role model. I think we could just accept any instant vector-typed result. If the expression doesn't evaluate to one or more info metric, we can issue a warning and not do anything. But generally, we are going down into details very prematurely here.

PR #40, opened 2024-11-26
Propose WAL format versioning and change strategy.

bwplotka:


proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,222 @@
...
+* TSDB format can but does not need to follow the same recommendations. We want to change the WAL format now, so prioritizing WAL.
+* Mentioning Write-Before-Log, it uses WAL format so all apply to WBL too.
+* To reduce scope we don't mention [memory snapshot format](https://github.com/prometheus/prometheus/blob/fd5ea4e0b590d57df1e2ef41d027f2a4f640a46c/tsdb/docs/format/memory_snapshot.md#L1) for now.
+
+## How
+
+We recommend the [Two-Fold Migration Strategy](#two-fold-migration-strategy) with one variation: a configuration flag that tells Prometheus to write a Y+1 version if the user desires.
+
+
+In other words, we propose to add a TSDB **integer** flag `--storage.tsdb.write-wal-format` that tells Prometheus to use a particular WAL format for both WAL and WBL. This kind of flag will change its default to a new version ONLY when the previous Prometheus version can read that version.
```
krajorama, line 85: I think we should be more specific about the version here. Prometheus defines LTS versions, so this should read 
```suggestion
In other words, we propose to add a TSDB **integer** flag `--storage.tsdb.write-wal-format` that tells Prometheus to use a particular WAL format for both WAL and WBL. This kind of flag will change its default to a new version ONLY when the previous LTS Prometheus version can read that version.
```

This would mean 2 years life time of a WAL version (but of course it may mean a jump of more than one version). So let's suppose we implement this flag and the default is  1 in version 3.1 LTS (or something) . To set the default to 2 (or more) we need an LTS release a year later that can read version 2, but defaults to 1. Another year later we could set default 2.

WDYT?

krajorama: oh, I later read the alternatives - I think it makes sense to instill a sense of stability

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,222 @@
...
+Pros:
+* This gives a bit more stability to users and less surprise.
+
+Cons:
+* We don't do LTS-es regularly
+* Is it safe or feasible to add Y+1 support to the old LTS code?
+
+### Add a CLI tool that rewrites WAL to a specific version
+
+We could add a `promtool` command that rewrites WAL segments to a given version.
```
krajorama, line 157: Since we have more automatic options, I would not do this.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,222 @@
...
+Pros:
+* Less need for a two-fold strategy?
+
+Cons:
+* We need to write migration code and ensure it's efficient enough for bigger data
+* A bit painful to use on scale and remotely (e.g. on Kubernetes)
+
+### Rewrite WAL before/during replay
+
+Instead of supporting multiple directories for WAL for various versions, we could rewrite WAL on the start.
```
krajorama, line 172: Cons+ I think the risk is that the rewrite fails and we cannot go back , resulting in data loss. Read only WAL on startup seems safer.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,222 @@
...
+### Require LTS support
+
+We could add a variation to the [Two-Fold Migration Strategy](#two-fold-migration-strategy) where both X+1 and the last LTS (long time support version) before X+1, should understand Y+1 version -- before X+2 can write it by default.
+
+Pros:
+* This gives a bit more stability to users and less surprise.
+
+Cons:
+* We don't do LTS-es regularly
+* Is it safe or feasible to add Y+1 support to the old LTS code?
```
krajorama, line 153: I think it will take us some time to actually have a good version 2. There's a bunch of things to implement, so it's not far fetched to introduce it in the next LTS.

krajorama: We could introduce the flag right now as experimental. Use at own risk :) In fact maybe it has to be "experimental" all the time since you don't guarantee forward compatibility. Which would mean that:

There is no need to *set* a version through a flag. You need the `--version` flag to print the default version and the experimental version.

And to use the experimental version you'd set the `--enable-feature  wal-vX` . This would imply that we can break anything while the next version is experimental.

proposals/2024-11-25_changing_wal_format.md
bwplotka: Ah, sorry. Copy-pasta. I really wanted to NOT do LTS for now, but initially I thought that's a good idea. This should be in alternatives. WDYT?

proposals/2024-11-25_changing_wal_format.md
bwplotka: Well, if we accept that automation == going through concrete releases when upgrading (kind of 💩  but a common DB practice)

proposals/2024-11-25_changing_wal_format.md
bwplotka: > I think it will take us some time to actually have a good version 2. 

I like the idea of not forcing ourselves to only +1 version compatibility, but do the compatibility (and opt-in) as long as it's needed.

> There is no need to set a version through a flag. You need the --version flag to print the default version and the experimental version.

Prometheus --version? What do you mean? It prints only Prometheus versions. 

> And to use the experimental version you'd set the --enable-feature  wal-vX . This would imply that we can break anything while the next version is experimental.

Fair, we could do that instead of special flag. However it would be a very special feature flag which means you can't really go back to certain Prometheus version. How to not surprise users with that?

bwplotka: I will rearrange proposal to switch into the ideas we discussed there, good points! I wonder if a new kind of feature flag would be useful here (e.g. `enable-storage-feature` or `enable-stateful-feature`).

bwplotka: Ok, I thought through this. I don't think feature flag makes sense here -- added alternative option in the doc on why. Can you check? Let's open a new discussion. I added more details on the fact that we can keep "forward compatible, but not wriiting version" for more than one release in the proposal 🤗 

Thanks, good points!

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,257 @@
...
+## How
+
+We recommend the [Two-Fold Migration Strategy](#two-fold-migration-strategy) with two details:
+
+* A new flag that tells Prometheus what WAL version to write.
+* There can be multiple "forward compatible" version, but the official minimum is one (see, the rejected [LTS idea](#require-lts-support))
+
+We propose to add a string `--storage.tsdb.stateful.write-wal-version` flag, with the default to `v1` that has a "stateful" consequence -- once new version is used, users will be able to revert only to certain Prometheus versions. Help of this flag will explain clearly what's possible and what Prometheus version you will be able to revert to.
+ 
+In other words, we propose to add a TSDB flag `--storage.tsdb.stateful.write-wal-version=(v1 or v2)` that tells Prometheus to use a particular WAL format for both WAL and WBL. This kind of flag will change its default to a new version ONLY when (at least) one previous Prometheus version can read that version (while writing the old one).
```
krajorama, line 93: I've never seen such notation, looks weird, let's use what you get when you run `prometheus --help`.
```suggestion
In other words, we propose to add a TSDB flag `--storage.tsdb.stateful.write-wal-version=<version>` that tells Prometheus to use a particular WAL format for both WAL and WBL. This kind of flag will change its default to a new version ONLY when (at least) one previous Prometheus version can read that version (while writing the old one). The initial version would be `v1`.
```

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,257 @@
...
+* Users have a durable rollout path back and forth.
+* Dev has clarity on how to develop "breaking revert" changes.
+* In theory, it allows revert to X from X+2, by going through X+1 and ensuring all data was migrated (eventually) to Y version. In practice however that eventuality is long, or hard to discover.
+
+Cons:
+* It can catch users by surprise
+  * **Mitigation**: We also plan guide and documentation.
+* It takes time to rollout new changes if you want them fast.
+  * **Mitigation**: We plan to add a flag to opt-in sooner
+* It is a breaking change which shouldn't be made in a major release.
```
krajorama, line 151: Did you mean minor?
Also, is changing the default write format really a breaking change? The user has to do nothing and they can downgrade. The breaking change comes when you remove support for version Y.

I think the way around deferring removal of the old version Y to major release is to introduce a deprecation policy. Set Y+1 as default, deprecate the old version and mark it in docs/startup message. Then remove after 2 minor versions or more.
This would only affect people that actually set Y as explicit version.

```suggestion
* It is a breaking change which shouldn't be made in a minor release.
```

proposals/2024-11-25_changing_wal_format.md
bwplotka: That's what I meant (: thx.

proposals/2024-11-25_changing_wal_format.md
bwplotka: Yes I meant minor. 

Or we could try to keep it until the next major version WDYT? I think I would propose to not set it in stone - depends how easy is to maintain old format support.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+
+* dedicated guide for users
+* flag help
+* tsdb/docs for devs
+
+To achieve WAL versioning we also propose to start versioning the WAL, as [a whole format](https://github.com/prometheus/prometheus/blob/main/tsdb/docs/format/wal.md). This is necessary to communicate a breaking change and to tell Prometheus what WAL format to write in.
+
+We propose the addition of a `meta.json` file in the wal directory, similar to [block meta.json](https://github.com/prometheus/prometheus/blob/5e124cf4f2b9467e4ae1c679840005e727efd599/tsdb/block.go#L171), with Version field set to `1` for the current format and `2` for new changes e.g. when we start to write [new records](https://github.com/prometheus/prometheus/pull/15467/files). No `meta.json` is equivalent to `{"version":1} `meta.json` file.
+
+We propose to also store the new WALs in separate directories e.g. `wal.v2`. Thanks to that the rewrite from one version to another is **eventual** and can be done segment by segment without any forcible rewrite. The WAL will be rewritten within the next 2h of a normal operation. The additional advantage of this way of versioning is that it's clear when your WAL fully migrated to a certain version.
```
bboreham, line 114: Should mention checkpoints here.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+Instead of supporting multiple directories for WAL for various versions, we could rewrite WAL on the start.
+
+Pros:
+* Simpler implementation
+
+Cons:
+* We already suffer from replay problems, so I propose an eventual rewrite.
+* Rewrite is more risking than read-only WAL (of previous version)
+
+### Don't version WAL, don't introduce a flag
```
bboreham, line 200:
```suggestion
### Alternative: Don't version WAL, don't introduce a flag
```
I suggest to put "Alternative" in each of these sub-headings, because it isn't clear once you've scrolled down a bit.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+
+Cons:
+* It can catch users by surprise.
+  * **Mitigation**: We also plan guide and documentation.
+* It takes time to rollout new changes if you want them fast.
+  * **Mitigation**: We plan to add a flag to opt-in sooner.
+
+## Alternatives
+
+### Require LTS support
```
bboreham, line 154: This is more of an extension than an alternative.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+
+Cons:
+* Takes more time for features to be available to users
+* Demotivating for format changes (long feedback loop)
+* Harder to communicate what exactly changed in each Prometheus version or even implement backward compatibility?
+
+### Record-based or segment based WAL versioning
+
+Given we usually change WAL by changing its records, the WAL version could be simply [max number of types](https://github.com/prometheus/prometheus/blob/5e124cf4f2b9467e4ae1c679840005e727efd599/tsdb/record/record.go#L54) we write to WAL.
+Alternatively, it could be per segment e.g. introduce a special version record type that is only in the front of the segment file.
```
bboreham, line 213: Mention checkpoints here.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+Pros:
+* Simpler to implement now.
+* This would allow mixing Y and Y+1 segments/records in the same WAL, we would not need a new directory.
+
+Cons:
+* Won't work for bigger changes e.g.
+  * changes that merge records 
+  * sharding?
+
+### Maintain two WALs (well four, with WBL)
```
bboreham, line 224: why do we have separate WAL and WBL ?

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+* TSDB format can but does not need to follow the same recommendations. We want to change the WAL format now, so prioritizing WAL.
+* Mentioning Write-Before-Log, it uses WAL format so all apply to WBL too.
+* To reduce scope we don't mention [memory snapshot format](https://github.com/prometheus/prometheus/blob/fd5ea4e0b590d57df1e2ef41d027f2a4f640a46c/tsdb/docs/format/memory_snapshot.md#L1) for now.
+
+## How
+
+We recommend the [Two-Fold Migration Strategy](#two-fold-migration-strategy) with two details:
+
+* A new flag that tells Prometheus what WAL version to write.
+* There can be multiple "forward compatible" version, but the official minimum is one (see, the rejected [LTS idea](#require-lts-support))
```
bboreham, line 89: Who rejected it?

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+4. 3.25 can now switch to WAL v2.
+5. 4.0 can remove WAL v1 support.
+
+Pros:
+* Gives a bit more stability to users and less surprises.
+
+Cons:
+* Extremely heavy process that will make us afraid/refuse to make improvements to WAL, because it's too much work. It might fails our goal of `Balancing development velocity with user data stability risks`
+  * One mitigation would be an LTS retroactive strategy e.g. LTS 3.1 only supports WALv1, 3.3 adds WALv2, we do 3.1.1 with WALv2 too, 3.4 can now switch to WALv2, 4.0 can remove WALv1 support. It gives us more flexibility, but it's not very realistic to do patch release of LTS with risky feature like a new WAL.
+* We literally have no formal process for LTS versions and we don't do them regularly.
```
bboreham, line 172:
```suggestion
* We have no formal process for LTS versions and we don't do them regularly.
```

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+
+We could add a variation to the [Two-Fold Migration Strategy](#two-fold-migration-strategy) (let's call it a "LTS migration strategy") where both X+1 and the last LTS (long time support version) before X+1, is able to read Y+1 version. Only then we are allowed to release X+2 that switches the default.
+
+For example:
+
+1. LTS 3.1 only supports WAL v1.
+2. 3.3 adds WAL v2.
+3. we wait unit next LTS so e.g. 3.24.
+4. 3.25 can now switch to WAL v2.
+5. 4.0 can remove WAL v1 support.
```
bboreham, line 164: I think you should write down what the upgrade path looks like if this approach is not chosen, in the main section of this document.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+## How
+
+We recommend the [Two-Fold Migration Strategy](#two-fold-migration-strategy) with two details:
+
+* A new flag that tells Prometheus what WAL version to write.
+* There can be multiple "forward compatible" version, but the official minimum is one (see, the rejected [LTS idea](#require-lts-support))
+
+We propose to add a string `--storage.tsdb.stateful.write-wal-version` flag, with the default to `v1` that has a "stateful" consequence -- once new version is used, users will be able to revert only to certain Prometheus versions. Help of this flag will explain clearly what's possible and what Prometheus version you will be able to revert to.
+ 
+In other words, we propose to add a TSDB flag `--storage.tsdb.stateful.write-wal-version=<version>` that tells Prometheus to use a particular WAL format for both WAL and WBL. This kind of flag will change its default to a new version ONLY when (at least) one previous Prometheus version can read that version (while writing the old one). The initial version would be `v1`.
```
bboreham, line 93: I don't think one version (6 weeks) is enough to say "v2 is good enough, turn it on for everyone".

proposals/2024-11-25_changing_wal_format.md
bboreham: Yes, one year https://prometheus.io/docs/introduction/release-cycle/

proposals/2024-11-25_changing_wal_format.md
bwplotka: E_TOO_MANY_ALTERNATIVES

proposals/2024-11-25_changing_wal_format.md
bwplotka: I propose it to reject in this proposal. If we decide opposite, will switch. Sorry for confusion, I assume I can't use past lang here?

proposals/2024-11-25_changing_wal_format.md
bwplotka: It's written in the Two-Fold strategy section, but I can use the same versions here.

proposals/2024-11-25_changing_wal_format.md
bwplotka: I think I would love to keep the gate open if the change will be trivial (like literally we have this case now with nhcb IMO, but we can argue) (:

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+See [alternatives](#alternatives) for other ideas.
+
+### Two-Fold Migration Strategy
+
+Given the following example:
+
+![twofold.png](../assets/2024-11-25_changing_wal_format/twofold.png)
+
+1. We release Prometheus X+1 version that supports both Y and Y+1 data but still writes Y.
+2. We release Prometheus X+2 version that supports both Y and Y+1 data, but now it writes new data as Y.
```
carrieedwards, line 136:
```
2. We release Prometheus X+2 version that supports both Y and Y+1 data, but now it writes new data as Y.
```
For clarity: does this mean that the new record type (Y+1) will be re-named to Y?

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,260 @@
...
+2. 3.3 adds WAL v2.
+3. we wait unit next LTS so e.g. 3.24.
+4. 3.25 can now switch to WAL v2.
+5. 4.0 can remove WAL v1 support.
+
+Pros:
+* Gives a bit more stability to users and less surprises.
+
+Cons:
+* Extremely heavy process that will make us afraid/refuse to make improvements to WAL, because it's too much work. It might fails our goal of `Balancing development velocity with user data stability risks`
```
carrieedwards, line 170:
```suggestion
* Extremely heavy process that will make us afraid/refuse to make improvements to WAL, because it's too much work. It might fail our goal of `Balancing development velocity with user data stability risks`
```

proposals/2024-11-25_changing_wal_format.md
carrieedwards: Should we have some sort of distinction in the policy for when the flag's default value can be changed for cases where the WAL change only impacts an experimental feature (such as NHCB)?

proposals/2024-11-25_changing_wal_format.md
bwplotka: Because WBL is a separate WAL on disk

proposals/2024-11-25_changing_wal_format.md
bboreham: Yes, this would extend what we said we would do in LTS releases. 
The risk is something to weigh up.

proposals/2024-11-25_changing_wal_format.md
bboreham: That's a re-statement of the fact.  I am asking what the reason behind it is.

proposals/2024-11-25_changing_wal_format.md
bwplotka: removed this wording

proposals/2024-11-25_changing_wal_format.md
bwplotka: Also added https://prometheus.io/docs/introduction/release-cycle/ mention, thanks!

proposals/2024-11-25_changing_wal_format.md
bwplotka: Ups should be ` We release Prometheus X+2 version that supports both Y and Y+1 data, but now it writes new data as Y+1.`

Thanks for catching!

bwplotka:
Ok, so from the comments so far and [slack](https://cloud-native.slack.com/archives/C082ALTBY4S/p1733129928553079) seems to me like there is a sentiment to:

1) Switch to versioning segments, not directories.
2) Force last LTS to support the new version BEFORE changing the default to write new version; consider backports if the risk is small.

WDYT @krajorama @bboreham should I switch my proposal to those?

krajorama:
> Ok, so from the comments so far and [slack](https://cloud-native.slack.com/archives/C082ALTBY4S/p1733129928553079) seems to me like there is a sentiment to:
> 
>     1. Switch to versioning segments, not directories.
Yes, segments already provide ordering and would help us in upgrade->downgrade->upgrade scenarios. 
On a practical side, segments don't have a header, so I don't know how we'd do it, but that's another question.

> 
>     2. Force last LTS to support the new version BEFORE changing the default to write new version; consider backports if the risk is small.
Yes. I think there's a class of user that would very much expect this.

> 
> 
> WDYT @krajorama @bboreham should I switch my proposal to those?

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,281 @@
+## Recommendations for WAL Format Changes
+
+* **Owners:**
+  * [@bwplotka](https://github.com/bwplotka)
+
+* **Contributors:**
+  * [@krajorama](https://github.com/krajorama)
+  * [@bboreham](https://github.com/bboreham)
+  * [carrieedwards](https://github.com/carrieedwards)
```
nicolastakashi, line 9:
```suggestion
  * [@carrieedwards](https://github.com/carrieedwards)
```

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,324 @@
...
+* TSDB format can, but does not need to follow the same recommendations. We want to change the WAL format now, so prioritizing WAL.
+* Mentioning Write-Before-Log (WBL) or checkpoints, both uses WAL format internally.
+* To reduce the scope we don't mention [memory snapshot format](https://github.com/prometheus/prometheus/blob/fd5ea4e0b590d57df1e2ef41d027f2a4f640a46c/tsdb/docs/format/memory_snapshot.md#L1) for now.
+
+## How
+
+We recommend the [Two-Fold Migration Strategy](#how-two-fold-migration-strategy) with the two important additions:
+
+* A new flag that tells Prometheus what WAL version to write.
+* There can be multiple "forward compatible" versions, but the official minimum is **the previous [Long-Term-Support (LTS) release](https://prometheus.io/docs/introduction/release-cycle/)**. The forward compatibility can be optionally backported to the LTS, instead of waiting up to a 1y, depending on the risk.
```
codesome, line 99: If we backport the forward compatibility to the last LTS, will we also bump the default write version in the upcoming release before the next LTS?

If no: then why backport in the first place.

If yes: what about other "breaking" features that are in the latest releases that is not in the last LTS - you can't really rollback to the last LTS anymore because of that. Maybe this is the first step in the long term goal of all breaking additions can have forward compatibility in the last LTS, but I don't see it practically happening soon (maybe we enjoy this phase until we decide to break some other part of TSDB).

codesome (commented):
Thanks for writing this! I just have a few questions for clarification

proposals/2024-11-25_changing_wal_format.md
codesome: When implementing out-of-order, the WAL had all the out-of-order samples in them (both the ingested and uningested out-of-order samples). So we needed a way to know exactly what out-of-order samples were ingested, so that we can restore only those samples. We could only know if a sample went into a chunk at the time of writing to the chunk (i.e. after writing to WAL), so we introduced WBL to log the out-of-order samples ingested, since it was logged after samples went into the chunk. Data in WBL only goes into the out-of-order head.

codesome: This could probably be merged together by writing separate OOO records in WAL. But the replay needs to be changed carefully so that things don't break.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,324 @@
...
+
+* [record type](https://github.com/prometheus/prometheus/blob/e410a215fbe89b67f0e8edef9de25ede503ea4e0/tsdb/record/record.go#L38)
+* [metric type in metadata record](https://github.com/prometheus/prometheus/blob/e410a215fbe89b67f0e8edef9de25ede503ea4e0/tsdb/record/record.go#L111)
+* [metadata fields are arbitrary labels in metadata record](https://github.com/prometheus/prometheus/blob/e410a215fbe89b67f0e8edef9de25ede503ea4e0/tsdb/record/record.go#L608).
+
+Historically, we didn't hit major problems because we were only adding new semantic data (e.g. exemplars, metadata, new native histograms) as new records. 
+However, these days, we need to add features to existing data (e.g. custom buckets that will replace classic histograms, cleanup of histogram records or created timestamps to samples). Even if we create a new record for those and use it for new samples, any rollback will **lose that information as they appear unknown in the old version**.
```
machine424, line 55: I don't see how we can do better than `Even if we create a new record for those and use it for new samples, any rollback will **lose that information as they appear unknown in the old version**`, even with the proposed changes, an old Prometheus version cannot do better than treating the new format data/records as unknown. as part of the forward compatibility.
So, in all cases, the old Prometheus version will lose that information as it doesn't know how to interpret it.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,324 @@
...
+
+Pro:
+* One issue with a flag is that we lose "automatic" rollout for some users e.g. what if we will have another v3 WAL version, but users sets an explicit `--storage.tsdb.stateful.write-wal-version=v3` option? For those users they will need to make a manual choice and there might be not aware of benefits of a new WAL or do any work to test new WAL for their setup. With raw two-fold strategy format is forced, but also transparent (if without issues).
+
+Cons:
+* Takes more time for features to be available to users
+* Demotivating for format changes (long feedback loop)
+* Harder to communicate what exactly changed in each Prometheus version or even implement backward compatibility?
+
+### Alternative: Use feature flag instead
```
machine424, line 305: I reviewed the linked issues and efforts motivating the segments versioning, and I believe all of them can be managed with the two-fold strategy/migration and a feature flag, so maybe I'm missing sth or maybe more concrete scenarios are missing in the proposal.

Unless we intend to change the record's header format (which I think we can also do without having to verson the WAL segments), I think the current format is generic enough to handle all other changes. It's a map of `record type/code -> how to handle it`, I think we can easily customize that mapping (and improve it if needed) without having to version the segments.

For example, If I were to propose an optimized encoding for samples, I would add a `SampleOptimized` record type (with its reader/writer) and a `--feature-flag=optimized-wal-encoding` or sth to trigger reading and then writing in another version. With the two-fold strategy and proper documentation/assistance on potential downgrades [1], as suggested above, and on feature gate deactivation [2], I believe we could roll out and test that `SampleOptimized`. This approach aligns with how we've proceeded so far, as you mentioned.

While versioning WAL segments provides more visibility into the version in use, what does a `v3` segment version really mean to a user? Does it mean native histograms are supported? Does it mean optimized encoding is enabled? I think it would only add confusion.

Also, having such vague versions would create additional confusion and necessitate more synchronization between developers working concurrently on changes that require "incrementing the WAL segment version".

---

[1] In addition to the documentation, we could implement a built-in safeguard mechanism for downgrades: store the minimum downgrade version on disk and refuse to start up unless the file is manually deleted or another action is taken. 
Also, if we want to raise awareness about the "dangers of unsupported downgrading", we may need to reconsider how `record.Unknown` is handled.

[2] We could leverage this proposal to improve the discoverability of built-in feature gates (I think there is an issue for this somewhere).

machine424 (commented):
came across this, thanks for putting this together.
Left some comments.

proposals/2024-11-25_changing_wal_format.md
bwplotka: Thanks, fair points. Sounds like you propose a feature flags with feature names, which semantically might be close to "record versioning" instead of a single "segment version" proposed right now. It seems the main difference is organizing features into version umbrella or deliver them in separation. Definitely something to consider, let's start unpacking pros & cons.

> While versioning WAL segments provides more visibility into the version in use, what does a v3 segment version really mean to a user? Does it mean native histograms are supported? Does it mean optimized encoding is enabled? I think it would only add confusion. Also, having such vague versions would create additional confusion and necessitate more synchronization between developers working concurrently on changes that require "incrementing the WAL segment version".

Yes. It will means immediately nothing what it changes to user and that's on purpose. This versioning is optimized for telling users about important data compatibility and discovering what version of Prometheus supports what versions of WAL, that's it.

Imagine we change WAL and finally use proper `nativehistogram` records (instead of 4 different records now) as a `wal-nhcleanup` feature, then new sample with CT with `wal-ctsample` feature, finally let's say structured summary in future with `wal-summary` feature. Now you deploy new Prometheus version that (after a fold) defaults to `wal-nhcleanup`, then user manually opt-in into `wal-ctsample`  and  `wal-summary`  is still opt-out. Let's say user reverts to the previous Prometheus release it was running (e.g. LTS or let's say 3 versions back) and some data are completely missing because of WAL incompatibility. Finding what feature actually is incompatible and what version of Prometheus supports what features is not easy. Talking in segment versions is a bit more consistent and obvious when `ls`-ing your directory or logging on unknown segment version.

But.. maybe that user experience is not needed. Maybe asking those tricky features to be in LTS version before switching to opt-in would help (I would avoid having to wait for v4.x Prometheus etc), better docs could be added too.

machine424: > and some data are completely missing because of WAL incompatibility

As I mentioned earlier, segment versioning isn't a perfect solution for this issue. If a user downgrades to a Prometheus version that can't decode `v2` segments, the data within those segments will be lost. However, if users follow the recommended downgrade paths, the feature flag approach should be able to handle those scenarios without data loss.

> Finding what feature actually is incompatible and what version of Prometheus supports what features is not easy. Talking in segment versions is a bit more consistent and obvious when ls-ing your directory or logging on unknown segment version.

I think adding a new dimension (segment version) to the mix would just add to the confusion here, especially since users will still need to consult documentation of other tools to figure out the differences between segment `v3` and `v2`.

I think well-documented feature flags and a well-documented compatibility matrix should be enough. The most important thing, as you mentioned, is to provide a well-tested fallback (through the two-fold strategy).

So, what I think we could still do through this proposal is:

- Officially document the two-fold strategy, with examples and schemas.
- Start requiring such "invasive/critical" (or maybe all of them)  feature flags to:
    - Ensure a fallback/LTS version exists and is mentioned in the docs before shipping the "invasive/feature flag in write mode".
    - Explicitly document what happens when the feature flag is disabled in the relevant versions.
- Add the needed utils/infra to allow e2e tests concerning those scenarios, in addition to those e2e tests, the feature flag can always be tested on long running stage envs even though a fallback version isn't ready yet, as long as the user is aware of that...

krajorama: Sure we have feature flags already and we can document how people can use the feature and what they have to be aware of when turning off the feature. But after feature(s) are mature, you want to enable them by default. So let's suppose we enable a bunch of features in version X. Someone upgrades to X and it doesn't work for some reason, they need to downgrade to X-1. Which fails because they didn't know what feature flags to enable.

Ok, so we should  make X-1 version be able to detect what features were enabled in X and read those files/data automatically. I think using a version for that purpose is pretty straightforward. And people don't have to really know about it. If you downgrade one version , it would just work. If you downgrade more than one LTS version, it can print a sensible error message - if needed.

In general I feel like we're going to have and want to have a single main track for the stable WAL implementation. In other words not force users to chose which kind of WAL they want to have. And having a version in it is pretty clear cut, same as TSDB. Most users would never notice.

So to summarize:
- this proposal does not prohibit feature flags and trying things that way - we already have a place to document feature flags, that could explain the compatibility (or lack of) aspects.
- when we want to move mainline forward, I think versioning makes sense as per this proposal.

bwplotka:
I will try to address new comments but would love your look @gouthamve and @jesusvazquez !

proposals/2024-11-25_changing_wal_format.md
machine424: > Sure we have feature flags already and we can document how people can use the feature and what they have to be aware of when turning off the feature. But after feature(s) are mature, you want to enable them by default. So let's suppose we enable a bunch of features in version X. Someone upgrades to X and it doesn't work for some reason, they need to downgrade to X-1. Which fails because they didn't know what feature flags to enable.

We can continue assuming that feature flags only disappear on major versions, but even if we suppose they can disappear within a minor version, I think what that user really needs is keeping track and versioning their Prometheus config, if they're making changes on the fly to their Prometheus config without keeping  track of their changes, well... I don't think adding WAL versioning to the mix is the solution here, the problem is somewhere else.

Even with WAL versioning as proposed here, users will need to set up/deal with versioning related config, so even then the scenario you shared isn't avoided.

With enough doc regarding the feature flags and a way to track and version Prometheus configs (not that much to ask from users), I think that scenario can be avoided.

> In general I feel like we're going to have and want to have a single main track for the stable WAL implementation. In other words not force users to chose which kind of WAL they want to have. And having a version in it is pretty clear cut, same as TSDB. Most users would never notice.

As mentioned earlier, the existing approach, where the WAL is simply a bunch of bytes with feature flags dictating their interpretation, is versatile enough to cover most scenarios. Unless I see a concrete example showing a limitation, I don't think introducing WAL versioning will add anything other than complexity to users and to developers.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,324 @@
...
+
+* Agree on the official recommendations for **lossless WAL migrations** strategies for devs and users.
+* Balancing development velocity with user data stability risks.
+* Ability to develop new version across releases in `main` branch.
+
+## Non-Goals
+
+* TSDB format can, but does not need to follow the same recommendations. We want to change the WAL format now, so prioritizing WAL.
+* Mentioning Write-Before-Log (WBL) or checkpoints, both uses WAL format internally.
+* To reduce the scope we don't mention [memory snapshot format](https://github.com/prometheus/prometheus/blob/fd5ea4e0b590d57df1e2ef41d027f2a4f640a46c/tsdb/docs/format/memory_snapshot.md#L1) for now.
```
jesusvazquez, line 92: Another non goal seems to be that we don't want to unify WAL and WBL on this effort.

proposals/2024-11-25_changing_wal_format.md
```diff
@@ -0,0 +1,324 @@
...
+* Users have a durable rollout path back and forth.
+* Dev has clarity on how to develop "breaking revert" changes.
+* In theory, it allows revert to X from X+2, by going through X+1 and ensuring all data was migrated (eventually) to Y version. In practice however that eventuality is long, or hard to discover.
+
+Cons:
+* It can catch users by surprise.
+  * **Mitigation**: We also plan guide and documentation and support LTS release.
+* It takes time to rollout new changes if you want them fast.
+  * **Mitigation**: We plan to add a flag to opt-in sooner.
+
```
jesusvazquez, line 207: Would it be viable to extend the WAL replay so that if there are unknown records in it, probably because you've rolled back to an older version, we keep take them out of the WAL file (or just keep them on disk if we use the `-vX` alternative) and keep them there until the user rolls forward again to a version that can consume them?

bwplotka:
Nice discussed happened on Slack https://cloud-native.slack.com/archives/C082ALTBY4S/p1744098976846579

I think we agreed to relax this proposal a bit and use feature flags for similar strategy.

I am trying to convince @machine424 to take ownership of this proposal as I go on paternity leave soon (: 

[One thing to figure out is WHAT's the actual strategy for feature flag lifecycle. Can we change them to opt-out without major version of Prometheus? There are mixed interpretations.](https://cloud-native.slack.com/archives/C082ALTBY4S/p1744195017866119?thread_ts=1744098976.846579&cid=C082ALTBY4S)

PR #41, opened 2024-12-04
proposal: Prombench Custom Benchmarks

bwplotka:
Proposes implementation for https://github.com/prometheus/test-infra/issues/659

### DIFF
--- a/proposals/2024-12-04_prombench-custom-benchmarks.md
+++ b/proposals/2024-12-04_prombench-custom-benchmarks.md
@@ -7,7 +7,7 @@
   * [@krajorama](https://github.com/krajorama)
   * [@ArthurSens](https://github.com/ArthurSens)
   
-* **Implementation Status:** `Partially implemented`
+* **Implementation Status:** `Implemented`
 
 * **Related Issues and PRs:**
   * https://github.com/prometheus/test-infra/issues/659
@@ -185,4 +185,4 @@ Cons:
 * [X] Extend `make deploy` and `make clean` to support git pulling on demand and dynamic directory choice. (https://github.com/prometheus/test-infra/pull/812)
 * [X] Add docs (https://github.com/prometheus/test-infra/pull/812)
 * [X] Extend Prometheus prombench GH job (https://github.com/prometheus/prometheus/pull/15682)
-* [ ] Announce on -dev list.
+* [X] Announce on -dev list.
--- a/proposals/2024-12-04_prombench-custom-benchmarks.md
+++ b/proposals/2024-12-04_prombench-custom-benchmarks.md
@@ -2,14 +2,20 @@
 
 * **Owners:**
   * [@bwplotka](https://github.com/bwplotka)
-
-* **Implementation Status:** `Not implemented`
+ 
+* **Contributors:**
+  * [@krajorama](https://github.com/krajorama)
+  * [@ArthurSens](https://github.com/ArthurSens)
+  
+* **Implementation Status:** `Partially implemented`
 
 * **Related Issues and PRs:**
   * https://github.com/prometheus/test-infra/issues/659
+  * https://github.com/prometheus/prometheus/pull/15682
+  * https://github.com/prometheus/test-infra/pull/812
 
 
-> TL;DR: We propose a way for maintainers to deploy custom Prombench benchmarks via `/prombench <...> -bench.version <branch or commit SHA>` PR command. Benchmarks can be then customized via custom branches on https://github.com/prometheus/test-infra repo that modify [any benchmark-specific manifests](https://github.com/prometheus/test-infra/tree/master/prombench/manifests/prombench).
+> TL;DR: We propose a way for maintainers to deploy custom Prombench benchmarks. We propose adding two flags to `/prombench`: `--bench.version=<branch or @commit SHA>` and `--bench.directory=<name of the benchmark scenario directory>` to customize benchmark scenarios.
 
 ## Why
 
@@ -21,7 +27,7 @@ Standard benchmark is a great way to have a uniform understanding of the efficie
 
 In other words, ideally Prombench would support a way to perform a custom benchmark when triggered it via Pull Request in Prometheus repo e.g. `/prombench v3.0.0 <please run agent mode with only native histograms load, OTLP receiving and remote write 2.0 with experimental arrow format`.
  
-Non-exhausting list of custom benchmarks that would be epic to do, discussed among community (`A` and `B` are two Prometheus process benchmark always run):
+Non-exhausting list of custom benchmarks that would be epic to do, discussed among community:
 
 * Different configurations, e.g.:
   * Custom flag changes (especially [feature flags](https://prometheus.io/docs/prometheus/latest/feature_flags/))
@@ -49,7 +55,7 @@ Non-exhausting list of custom benchmarks that would be epic to do, discussed amo
 
 ### Pitfalls of the current solution
 
-Currently, there is no easy way to run macro-benchmark for new, non-standard configurations and setups (technically there is some limited and hacky way - through manual GKE cluster operations on prombench after it starts).
+Currently, there is no easy way to run macro-benchmark for both non-standard and standard configurations and setups.
 
 This causes:
 
@@ -59,6 +65,8 @@ This causes:
 * Wasted time for people recreating custom macro benchmarks on their own, which are harder to repro or trust.
 * Overloading standard scenario with too many or too custom features.
 
+NOTE: Technically there are some limited and hacky ways for custom scenarios e.g. through manual GKE cluster operations on prombench after it starts or by hardcoding new defaults or logic on the benchmarked PR. While possible, they are neither easy or clean ways.
+
 ## Goals
 
 * Keep the standard scenario healthy and useful for release gating.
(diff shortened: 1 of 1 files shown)


proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+We don't discuss here the following changes:
+
+* Ability to run benchmarks from the GitHub Issues.
+* Ability to run multiple benchmarks per one PR.
+
+## How
+
+We propose changing Prombench in two phases:
+
+### Ability to specify verion of the benchmark `/prombench <...> -bench.version <branch or commit SHA>`
```
krajorama, line 81:
```suggestion
### Ability to specify version of the benchmark `/prombench <...> -bench.version <branch or commit SHA>`
```

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+
+Standard benchmark is a great way to have a uniform understanding of the efficiency for the generic Prometheus setup, especially useful before Prometheus releases. However, with more contributors and Prometheus features, there is an increasing amount of reasons to perform special, perhaps one-off benchmarking scenarios that target specific feature flags, Prometheus modes, custom setups or to elevate certain issues.
+
+In other words, ideally Prombench would support a way to perform a custom benchmark when triggered it via Pull Request in Prometheus repo e.g. `/prombench v3.0.0 <please run agent mode with only native histograms load, OTLP receiving and remote write 2.0 with experimental arrow format`.
+ 
+Non-exhausting list of custom benchmarks that would be epic to do, discussed among community (`A` and `B` are two Prometheus process benchmark always run):
```
krajorama, line 24: The part about A and B doesn't make sense to me, did you mean to list them somewhere?

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+
+* Different configurations, e.g.:
+  * Custom flag changes (especially [feature flags](https://prometheus.io/docs/prometheus/latest/feature_flags/))
```
krajorama, line 27: Strictly speaking you can already do this by enabling CT on the PR branch in Prometheus and the compared branch. Did you mean to say that you want to measure scenarios where CT is extensively used? I guess that's the not so obvious problem as the test definition is in infra-test repo.

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+    * *Is my improvement to CT meaningful, for `created-timestamp-zero-ingestion` feature flag enabled?*
+  * Custom configuration file / different recording rule
```
krajorama, line 29: Same as above , this is already possible now. Did you mean to use more histograms in the test?

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+* Different environment, e.g. to elevate certain issue:
+    * only native histograms metrics
+    * only PromQL histogram_quantile queries
+    * bigger load
+    * smaller load
+    * long term blocks
+    * remote write failures
+
+* Compare two, same version Prometheus-es, across features or environments, e.g.:
+  * *How better or worse it is when we enable feature X?*
```
krajorama, line 42: same as above

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+  * agent mode vs tsdb mode for scrape and write
+  * is a new WAL format better than the old one?
+
+* Testing potential improvements to standard scenario, e.g.:
+  * [switching fake-webserver to avalanche (tuning the load)](https://github.com/prometheus/test-infra/issues/559)
+  * adding initial blocks
+
+### Pitfalls of the current solution
+
+Currently, there is no easy way to run macro-benchmark for new, non-standard configurations and setups (technically there is some limited and hacky way - through manual GKE cluster operations on prombench after it starts).
```
krajorama, line 52: Again I think it's more a problem of no easy way to customize the scrape/query load.

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+* Ability to run benchmarks from the GitHub Issues.
+* Ability to run multiple benchmarks per one PR.
+
+## How
+
+We propose changing Prombench in two phases:
+
+### Ability to specify verion of the benchmark `/prombench <...> -bench.version <branch or commit SHA>`
+
+First, we add the ability to deploy from a https://github.com/prometheus/test-infra branch or commit SHA.
```
krajorama, line 83: Do people that will use this feature have the ability to push branches to prometheus/test-infra?

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+  * number of replicas
+  * prometheus config, flags and recording rules
+  * remote write sink options (e.g introduce failures)
+* Prombench core framework is intact - no cluster resources can be modified this way.
+* Prometheus team can maintain a few official branches for common scenarios e.g. `agent`
+* Anybody can inspect what will be deployed by checking https://github.com/prometheus/test-infra repo.
+* Only team can change manifests (contributors will need to create PRs).
+
+Cons:
+
```
krajorama, line 109: + when reviewing a PR you have to also review the test code to trust the results from the test

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,152 @@
...
+* Allows feature comparison explained in [Why section](#why).
+
+Cons:
+* One can "malform" benchmark to run different versions which is likely not useful, but it will be visible in a commit.
+
+## Alternatives
+
+### Alternative: Carefully design and template a set of customization to standard scenarion in `main`
+
+One could add a big configuration file/flag surface to `/prombench` allowing to carefully change various aspects.
```
krajorama, line 133: note: I think over night it would be good to migrate features from branches to the standard test suite. As for example OOO and native histograms are more stable.

krajorama (commented):
Thank you for this. I agree in general, this seems like a good idea. Made some comments.

proposals/2024-12-04_prombench-custom-benchmarks.md
bwplotka: You could hardcore a new default sure, but is it really a clean and explicit way for custom scenarios?

proposals/2024-12-04_prombench-custom-benchmarks.md
bwplotka: Yes. And if there are contributors they need us to trigger prombench anyway, so we can create a branch and curate PR for it too. 

However, maybe we should allow arbitrary forks too as we have a gate in terms of maintainer triggering a benchmark? Should we allow non test-infra repo branches and commits? 🤔

proposals/2024-12-04_prombench-custom-benchmarks.md
bwplotka: agree

proposals/2024-12-04_prombench-custom-benchmarks.md
krajorama: I did prefix it by "strictly speaking". 

I'm  not sure which is more explicit. In the PR that you are testing you can say and have in the code diff the change in the default config. With the infra-test branch you'd have to make a branch compare to main to see what happens.

This is not a blocker.

proposals/2024-12-04_prombench-custom-benchmarks.md
krajorama: You're right about the recording rule.

proposals/2024-12-04_prombench-custom-benchmarks.md
krajorama: I'm ok with maintainer curated branches at first. I'd be a little bit afraid to allow forks. What would keep people from getting approved on a branch then changing it to mine bitcoin? 

Not even sure if this consideration belongs to "pros" or "cons" column :)

ArthurSens (commented):
One thing I liked very much about this proposal is its easy and fast implementation! The lack of custom benchmarks blocks a lot of work, so having **_something_** going quickly is nice.

When we discussed this in the past, I had something different in mind inspired by how [Prow](https://docs.prow.k8s.io/docs/overview/) parses comments to add extra functionality to plugins.

The idea is to improve `comment-monitor` with new commands:
 
```
/prombench main
/extra-args --enable-features=native-histograms,wal-records --web.enable-otlp-receiver
/with avalanche
/with agent-mode
```

* `extra-args` is a special command that adds extra command line flags to Prometheus; it can be used to add feature flags or "regular" commands
* `with` is used for more complex modifications, like replacing/updating the load generator or features that require changing several parts of Prometheus. For example, agent mode requires an extra flag and removing parts of the config file.


--- 

The main downside of the current proposal is that maintaining several branches for each kind of benchmarks seems like a lot of work. If we make a change in the main branch that should be added to all others, the number of PRs would get out of control 😬

bwplotka:
Nice @ArthurSens - one of my alternatives is similar to that, so essentially "prombench options for every variation" -  where you have some kind of custom config passed via `/prombench` or comment in general. I will add your syntax too to this proposal doc.

> This proposal is a lot harder though, not sure how much time that would take 😬

It's a good question, changing comment-monitor is easy, applying those options further down (and validating) it's bit harder but I like branching a bit more, because it unlocks almost all customizations vs carefully curating prombench options and debating which ones to support, which one you have (you have to document those) etc.

Also,  my proposal and yours are not conflicting. We could implement branching so we can experiment and do more complex variations, but since it's harder to maintain those branches over time, we could implement common "variation"  options in the prombench long term (so your more proposal). WDYT?

cc @bboreham @krajorama @jesusvazquez @codesome any preferences on "branching" vs "prombech options" so far?

bwplotka:
There is also a bit more "maintainable" path where instead of branches we do[ custom directories](https://github.com/prometheus/test-infra/blob/master/prombench/manifests/prombench/benchmar) on `main` and you can specify different directory than default one. Then we can even use kustomize to create variations or just copy - bit easier to discover what's possible and maintain 🤔  Will add as a 3rd option.

proposals/2024-12-04_prombench-custom-benchmarks.md
bwplotka: Clarified in the pittfalls 👍🏽

proposals/2024-12-04_prombench-custom-benchmarks.md
```diff
@@ -0,0 +1,188 @@
...
+
+We propose to add the `--bench.version=<branch|@commit>` flag to `/prombench` GH PR command, with the default value `master`. This flag will cause [the prombench GH action](https://github.com/prometheus/prometheus/blob/main/.github/workflows/prombench.yml) to pull specific commit SHA (if `--bench.version` value is prefixed with `@`) or branch from the https://github.com/prometheus/test-infra before deploying (or cleaning) test benchmark run. For the default `master` value, it will use the existing flow with the `docker://prominfra/prombench:master` image files.
+ 
+Here are an example steps to customize and run a customized benchmark with `--bench.version` flag.
+
+1. Create a new branch on https://github.com/prometheus/test-infra e.g. `benchmark/scenario1`.
+2. Modify this directory to your liking e.g. changing query load, metric load of advanced Prometheus configuration. It's also possible to make Prometheus deployments and versions exactly the same, but vary in a single configuration flag, for feature benchmarking.
```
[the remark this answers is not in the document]
bwplotka: Nice, you got me

ArthurSens (approved):
LGTM, we can proceed with the alternative in another proposal if we ever feel like it :)

PR #44, opened 2025-01-05
promql: add vanish function

fatpat:


beorn7:
I believe that this functionality can be easily emulated with existing primitives.

For example `present_over_time(some_metric[1h]) unless present_over_time` will return all metrics named `present_over_time` that have existed at some point during the last hour but do not exist now.

beorn7:
With no further discussion having happened in two months, I assumed my point above is valid. I'll close this, but whoever wants to chime in here, please do so. We can re-open as needed.

PR #45, opened 2025-02-18
add 2025-01-02_direct-io.md

machine424:
kick off PR https://github.com/prometheus/prometheus/pull/15365

### DIFF
--- a/proposals/2025-01-02_direct-io.md
+++ b/proposals/2025-01-02_direct-io.md
@@ -35,7 +35,8 @@ achieved through direct I/O are welcome.
 
 In addition to the issues mentioned above, sometimes the page cache generated during writes is not
 used at all. For instance, once chunks are written during compaction, they are opened via
-`mmap`, which renders the page cache produced during writing redundant and useless.
+`mmap`, which may render the page cache produced during writing redundant (on `OpenBSD` e.g.) or/and
+useless.
 
 ## Goals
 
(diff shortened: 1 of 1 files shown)


proposals/2025-01-02_direct-io.md
```diff
@@ -0,0 +1,166 @@
...
+
+For a look at the outcomes, the screenshot below shows a reduction of `20-50%` in page cache usage,
+as measured by the `container_memory_cache` metric while running Prombench on
+[Prometheus PR #15365](https://github.com/prometheus/prometheus/pull/15365). In this test, the
+direct I/O writer was used for writing chunks during compaction. No significant regressions or
+adverse effects on CPU or disk I/O were observed.
+
+![container_memory_cache](../assets/2025-01-02_direct-io/container_memory_cache.png)
+
+## Alternatives
```
dgl, line 154: Using `posix_fadvise` with `POSIX_FADV_DONTNEED` could be an option if it doesn't make sense to totally avoid the cache (there's also Linux specific memory management options like `MADV_COLD`, which could help for some of the container concerns to optimise which memory is reclaimed, although it wouldn't likely have as much user visible impact on the page cache).

I suspect those doesn't make sense for the TSDB but could be worth mentioning here, there might be other places in the ecosystem where those make more sense than `O_DIRECT`.

proposals/2025-01-02_direct-io.md
```diff
@@ -0,0 +1,166 @@
...
+might lead users or admins to make poor decisions, such as allocating more memory than necessary.
+Moreover, bypassing the cache when it is not required can help eliminate the overhead of data copying
+and reduce additional kernel overhead from reclaiming the cache. Thus, any performance improvements
+achieved through direct I/O are welcome.
+
+### Pitfalls of the Current Solution
+
+In addition to the issues mentioned above, sometimes the page cache generated during writes is not
+used at all. For instance, once chunks are written during compaction, they are opened via
+`mmap`, which renders the page cache produced during writing redundant and useless.
```
dgl, line 38: I'm not sure that's entirely true, Linux has a merged page + buffer cache: https://www.thomas-krenn.com/en/wiki/Linux_Page_Cache_Basics#Page_Cache_or_Buffer_Cache

(The exception being OpenBSD, see https://github.com/prometheus/prometheus/issues/8799)

machine424: You're right, I'll make that clearer.

proposals/2025-01-02_direct-io.md
machine424: Thnaks, I'll mention that.

machine424: Also, work on `RWF_UNCACHED` has resumed https://lore.kernel.org/linux-fsdevel/20241111234842.2024180-1-axboe@kernel.dk/T/#cluster-upstream-ci

Thinking about going with an `use-uncached-io` feature flag, give direct I/O a chance, then reconsider once `RWF_UNCACHED` is usable.

proposals/2025-01-02_direct-io.md
```diff
@@ -0,0 +1,181 @@
...
+
+* **Other Docs or Links:**
+  * [Slack Discussion](https://cloud-native.slack.com/archives/C01AUBA4PFE/p1726674665380109)
+
+> This effort aims to experiment with direct I/O to determine whether it can enhance user
+experience and provide performance improvements.
+
+## Why
+
+Probably due to a lack of [understanding of good taste](https://yarchive.net/comp/linux/o_direct.html).
```
bwplotka, line 19: What's the purpose of this mention? (:

machine424: just quoting @torvalds
```
Sadly, database people don't seem to have any understanding of good taste,
and various OS people end up usually just saying "Yes, Mr Oracle, I'll
open up any orifice I have for your pleasure".

			Linus
```

bboreham (approved):

PR #46, opened 2025-02-24
Improve phrasing

dgl:


bwplotka (approved):
...you need to feel it too though (: 

Thanks!

PR #47, opened 2025-02-24
proposal: Support secret providers

hsmatulisgoogle:
This PR adds a proposal to support specifying secrets through secret providers.

TL;DR from doc
> TL;DR: This document proposes adding a new way for Prometheus to discover and use secrets from various secret providers, similar to how service discovery works. It introduces a new configuration section where users can specify different secret providers and their configurations. It also defines interfaces and methods for secret providers to implement, allowing for flexibility in how secrets are fetched and managed.

### DIFF
--- a/proposals/2025-02-24_secret_providers.md
+++ b/proposals/2025-02-24_secret_providers.md
@@ -1,0 +1,182 @@
+## Remote Secrets (Secret Providers)
+
+* **Owners:**
+  * Henrique Matulis (@hsmatulisgoogle)
+
+* **Implementation Status:** Not implemented
+
+* **Related Issues and PRs:**
+  * https://github.com/prometheus/prometheus/issues/8551
+  * https://github.com/prometheus/prometheus/issues/11477
+  * https://github.com/prometheus/alertmanager/issues/3108
+  * https://github.com/prometheus/prometheus/pull/13955
+  * https://github.com/prometheus/exporter-toolkit/pull/141
+  * https://github.com/prometheus/prometheus/issues/5795
+
+
+
+
+* **Other docs or links:**
+  * [Prometheus Remote Secrets Doc](https://docs.google.com/document/d/1EqHd2EwQxf9SYD8-gl3sgkwaU6A10GhiN7aw-2kx7NU/edit?tab=t.0)
+    * Previous proposal by @TheSpiritXIII
+  * https://stackoverflow.com/questions/43609144/can-prometheus-store-basic-auth-passwords-in-any-format-other-then-plain-text
+  * https://groups.google.com/g/prometheus-users/c/yWLE9qoG5GU/m/ke8ewxjIAQAJ
+  
+
+> TL;DR: This document proposes adding a new way for Prometheus to discover and use secrets from various secret providers, similar to how service discovery works. It introduces a new configuration section where users can specify different secret providers and their configurations. It also defines interfaces and methods for secret providers to implement, allowing for flexibility in how secrets are fetched and managed.
+
+## Why
+
+The motivation behind this design document is to enhance the security and flexibility of secret management in Prometheus. Currently, Prometheus only supports reading secrets from the filesystem or directly from the configuration file, which can lead to security vulnerabilities and limitations when working with certain service providers.
+
+This proposal introduces secret discovery, similar to service discovery, where different secret providers can contribute code to read secrets from their respective APIs. This would allow for more secure and dynamic secret retrieval, eliminating the need to store secrets in the filesystem and reducing the potential for unauthorized access.
+
+### Pitfalls of the current solution
+
+Storing secrets in the filesystem poses risks, especially in environments like Kubernetes, where any pod on a node can access files mounted on that node. This could expose secrets to attackers. Additionally, configuring secrets through the filesystem often requires extra setup steps in some enviroments, which can be cumbersome for users.
+
+Storing secrets inline can also pose risks, as the configuration file may still be accessible through the filesystem. Additionally it can lead to configuration files becoming cluttered and difficult to manage.
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Allow Prometheus to read secrets remotely from secret providers.
+* Introduce secret discovery, similar to service discovery, where different secret providers can contribute code to read secrets from their respective API.
+
+### Audience
+
+* Prometheus maintainers
+* Alertmanager maintainers
+* Secret providers interested in contributing code
+* Users looking to use secret providers
+
+## Non-Goals
+
+* Implement a variety of secret providers.
+
+## How
+
+
+### Interfaces
+
+Secret providers will be created from their configurations through the following interface:
+
+```
+type SecretProviderConfiguration interface {
+	// Returns the secret provider for the given configuration.
+	Load() (SecretProvider, error)
+}
+```
+
+Secret providers will have to satisfy the following interface. Secret providers will be expected to be long lived to allow for caching. However they will be re-instantiated in case of a configuration change. The Fetch method will be called before each http request done through `prometheus/common/config`
+
+```
+type SecretProvider interface {
+	// Returns the secret value for the given configuration.
+  Fetch(ctx context.Context, secretRef string) (string, error)
+}
+```
+
+### Configuration
+
+Globally there will be a section to configure secrete providers to be used throught the config file. Here is an example:
+
+```
+global:
+  secret_providers:
+  - name: my_secret_provider
+    kubernetes_sp_config:
+      namespace: ns1
+```
+
+
+For secret related fields under http_config, a new `ref` variant will be added that can reference these secret providers. For instance, basic_auth will have the following form
+
+```
+basic_auth:
+...
+  password: <secret>
+  password_file: <string>
+  password_ref:
+    <string>: <string>
+```
+
+and could be instantiated as follows:
+
+```
+basic_auth:
+  password_ref:
+    my_secret_provider: 'my-secret-key'
+```
+
+#### Full configuration example
+```
+global:
+  secret_providers:
+  - name: kube1
+    kubernetes_sp_config:
+      namespace: ns1
+  - name: kube2
+    kubernetes_sp_config:
+      namespace: ns2
+...
+  scrape_configs:
+  - job_name: 'http-basic-auth-endpoint'
+    http_config:
+      basic_auth:
+        username: 'myuser'
+        password_ref:
+          kube1: 'myuser-pass'
+    static_configs:
+    - targets: ['www.endpoint.com/basic-auth']
+  scrape_configs:
+  - job_name: 'http-authorization-auth-endpoint'
+    http_config:
+      authorization:
+        credentials_ref:
+          kube2: 'header-credentials'
+    static_configs:
+    - targets: ['www.endpoint.com/authorization-auth']
+  scrape_configs:
+  - job_name: 'tls-certificate-endpoint'
+    http_config:
+      tls_config:
+        key_ref:
+          kube2: 'header-credentials'
+    static_configs:
+    - targets: ['www.endpoint.com/tls-certificate']
+```
+
+
+## Alternatives
+
+### Modify Secret type
+
+Currently most secrets in the config use the prometheus.common.config.Secret alias. We could modify this type such that if only a string is passed it behaves the same as before, and if a map is passed in it assumes it to be a reference to be fetched from the associated secret provider.
+
+This would mean the config would look like this instead:
+```
+basic_auth:
+  password: 'my-secret-value'
+...
+basic_auth:
+  password:
+    my_secret_provider: 'my-secret-key'
+```
+
+Pros:
+* Simpler, more unified config
+* Defines a clear way to use secrets outside of the `http_config` component
+
+Cons:
+* Can be more confusing
+* Requires more careful documentation
+* Unclear how to refresh the configs for arbitrary components that could now change dynamically
+
+
+## Action Plan
+
+* [x] Add a secret manager to `prometheus.common.config`
+* [ ] Add secret providers to prometheus  
+* [ ] Add docs 


proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+* **Related Issues and PRs:**
+  * https://github.com/prometheus/prometheus/issues/8551
+  * https://github.com/prometheus/prometheus/issues/11477
+  * https://github.com/prometheus/alertmanager/issues/3108
+  * https://github.com/prometheus/prometheus/pull/13955
+  * https://github.com/prometheus/exporter-toolkit/pull/141
+  * https://github.com/prometheus/prometheus/issues/5795
+
+
+
```
bwplotka, line 17:
```suggestion
```

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+	Load() (SecretProvider, error)
+}
+```
+
+Secret providers will have to satisfy the following interface. Secret providers will be expected to be long lived to allow for caching. However they will be re-instantiated in case of a configuration change. The Fetch method will be called before each http request done through `prometheus/common/config`
+
+```
+type SecretProvider interface {
+	// Returns the secret value for the given configuration.
+  Fetch(ctx context.Context, secretRef string) (string, error)
```
bwplotka, line 77:
```suggestion
	// Returns the secret value for the given configuration.
	Fetch(ctx context.Context, secretRef string) (string, error)
```

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+
+For secret related fields under http_config, a new `ref` variant will be added that can reference these secret providers. For instance, basic_auth will have the following form
+
+```
+basic_auth:
+...
+  password: <secret>
+  password_file: <string>
+  password_ref:
+    <string>: <string>
```
bwplotka, line 102: What's the format of key and value here?

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+    static_configs:
+    - targets: ['www.endpoint.com/tls-certificate']
+```
+
+
+## Alternatives
+
+### Modify Secret type
+
+Currently most secrets in the config use the prometheus.common.config.Secret alias. We could modify this type such that if only a string is passed it behaves the same as before, and if a map is passed in it assumes it to be a reference to be fetched from the associated secret provider.
```
bwplotka, line 156: Sounds like this alternative proposes a way to reuse `password` key (instead of adding new `password_ref`). I think this is very good to discuss and decide what's better 👍🏽 

(I would make this explicit and reduce mentions of code interfaces for now)

bwplotka (commented):
Great progress thanks!

proposals/2025-02-24_secret_providers.md
dgl: I like it too, but we have to be careful -- a loosely typed field like that can't (nicely) be represented in some schemas such as a Kubernetes CRD. I got strong pushback in the past when I proposed using something like this in just the unit test format.

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+    kubernetes_sp_config:
+      namespace: ns2
+...
+  scrape_configs:
+  - job_name: 'http-basic-auth-endpoint'
+    http_config:
+      basic_auth:
+        username: 'myuser'
+        password_ref:
+          kube1: 'myuser-pass'
```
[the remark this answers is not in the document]
rajagopalanand:
```
global:
  secret_providers:
  - name: kube1
    kubernetes_sp_config:
      namespace: ns1
  - name: kube2
    kubernetes_sp_config:
      namespace: ns2
```
Looks like users can define multiple secret providers to be instantiated. 

Consider the following

```
global:
  secret_providers:
  - kubernetes_sp_config:
      secrets:
      - name: kube1
        namespace: ns1
      - name: kube2
        namespace: ns2
  - aws_secret_manager_config:
      secrets:
      - name: secret1
        arn: arn:aws:secretsmanager:us-west-2:{account_id}}:secret:receiver-pager-duty-abcdef
      - name: secret2
        arn: arn:aws:secretsmanager:us-west-2:{account_id}}:secret:receiver-pager-duty-ghijk
        kms_key_arn: {arn of KMS key}
```
In this, users just define secret providers, tell it what secrets and configurations related to the secrets. 

Apologies if I misunderstood this

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+```
+basic_auth:
+  password_ref:
+    my_secret_provider: 'my-secret-key'
+```
+
+#### Full configuration example
+```
+global:
+  secret_providers:
```
rajagopalanand, line 116:
```
global:
  secret_providers:
  - name: kube1
    kubernetes_sp_config:
      namespace: ns1
  - name: kube2
    kubernetes_sp_config:
      namespace: ns2
```
With the above config, it looks like users can define multiple instances of the same secret provider to be instantiated. 

Consider the following

```
global:
  secret_providers:
  - kubernetes_sp_config:
      secrets:
      - name: kube1
        namespace: ns1
      - name: kube2
        namespace: ns2
  - aws_secret_manager_sp_config:
      secrets:
      - name: secret1
         secret_arn: arn:aws:secretsmanager:us-west-2:{{account_id}}:secret:receiver-pager-duty-abcdef
      - name: secret2
         secret_arn: arn:aws:secretsmanager:us-west-2:{{account_id}}:secret:receiver-pager-duty-ghijk
         kms_key_arn: {arn of KMS key}
```

In this, users just define which secret providers to use, tell it what secrets and configurations related to the secrets. In this format, each secret provider is defined only once

Apologies if I misunderstood this

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
yeya24: Agree. From the example it is unclear to me if `myuser-pass` is the name of the k8s secret itself or the secret key ref of one secret. 

A single primary key might be confusing and we lose some sort of `schema` if we gonna add new things to the primary key.
I feel something more complex makes sense.

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+Secret providers will be created from their configurations through the following interface:
+
+```
+type SecretProviderConfiguration interface {
+	// Returns the secret provider for the given configuration.
+	Load() (SecretProvider, error)
+}
+```
+
+Secret providers will have to satisfy the following interface. Secret providers will be expected to be long lived to allow for caching. However they will be re-instantiated in case of a configuration change. The Fetch method will be called before each http request done through `prometheus/common/config`
```
rapphil, line 72: >The Fetch method will be called before each http request done through `prometheus/common/config`

Can you clarify what the above statement means? Also what is secretRef?

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+* Prometheus maintainers
+* Alertmanager maintainers
+* Secret providers interested in contributing code
+* Users looking to use secret providers
+
+## Non-Goals
+
+* Implement a variety of secret providers.
+
+## How
```
rapphil, line 58: nit: reading through the proposal, it is simpler to understand if you present the configuration file first and then talk about the implementation later.

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Should be fixed!

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: N/A in new draft

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Both issues should be mentioned in the new draft

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Moved it around to be the main proposal and switched to suggest yaml tags

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+Pros:
+* Simpler, more unified config
+* Defines a clear way to use secrets outside of the `http_config` component
+
+Cons:
+* Can be more confusing
+* Requires more careful documentation
+* Unclear how to refresh the configs for arbitrary components that could now change dynamically
+
+
```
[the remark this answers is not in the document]
hsmatulisgoogle: I added a new section for the repo discussion but it is still missing the trade off analysis

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+```
+type SecretProvider interface {
+	// Returns the secret value for the given configuration.
+  Fetch(ctx context.Context, secretRef string) (string, error)
+}
+```
+
+### Configuration
+
+Globally there will be a section to configure secrete providers to be used throught the config file. Here is an example:
```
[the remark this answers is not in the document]
hsmatulisgoogle: Thanks, sorry for all the typos

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
hsmatulisgoogle: I switched the proposal to use the yaml tags, so there can be more precise naming per provider. Does this help with this issue?

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,182 @@
...
+    my_secret_provider: 'my-secret-key'
+```
+
+#### Full configuration example
+```
+global:
+  secret_providers:
+  - name: kube1
+    kubernetes_sp_config:
+      namespace: ns1
```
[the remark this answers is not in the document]
hsmatulisgoogle: Thanks!

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: This can indeed be done, and in theory if you use many configurations for one provider it would save space vs if you use many providers with one configuration each it would take up more space. I am still thinking about this and need to include it at least in the trade off section still

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Sorry for the confusion, i got rid of the implementation details as it is confusing and not relevant to the larger design discussion for now

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Thanks, I restructured the doc to focus a bit more on the config file and removed some implementation details

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,173 @@
...
+        password: !secret id=myk8secrets secret_id=pass1
+...
+        password: !secret
+          id: myk8secrets
+          secret_id: pass2
+```
+
+### Inline secrets
+
+When specifying secrets inline in the config file, the inline provider will be used. If a string is passed in, it is automatically converted to an inline provider to be consistent with previous syntax.
```
rapphil, line 91: Can you explicitly state what are valid syntaxes users can use for the case a inline secret is used? Also, if they are converted under the hood, that is a implementation detail? do we need to expose in the implementation?

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,173 @@
...
+
+Storing secrets in the filesystem poses risks, especially in environments like Kubernetes, where any pod on a node can access files mounted on that node. This could expose secrets to attackers. Additionally, configuring secrets through the filesystem often requires extra setup steps in some environments, which can be cumbersome for users.
+
+Storing secrets inline can also pose risks, as the configuration file may still be accessible through the filesystem. Additionally it can lead to configuration files becoming cluttered and difficult to manage.
+
+## Goals
+
+Goals and use cases for the solution as proposed in [How](#how):
+
+* Allow Prometheus to read secrets remotely from secret providers.
```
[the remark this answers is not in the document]
hsmatulisgoogle: Thanks, done!

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: You are right, this is an implementation detail and has been removed

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: The global section to re-use secret provider configs has been left to a later proposal, so this no longer applies

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,136 @@
...
+Wherever a `<secret>` type is present in the configuration files, we will allow users to specify a map specifying how to fetch from a secret provider.
+
+```
+        password:
+          provider: kubernetes
+          namespace: ns1
+          secret_id: pass2
+```
+
+### Inline secrets
```
rapphil, line 76:
```suggestion
### Backward Compatibility
```

Maybe change the name of the section to make it explicit that this is nothing new introduced by this proposal.

Also do we need to add a goal about backwards compatibility?

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,136 @@
...
+
+Wherever a `<secret>` type is present in the configuration files, we will allow users to specify a map specifying how to fetch from a secret provider.
+
+```
+        password:
+          provider: kubernetes
+          namespace: ns1
+          secret_id: pass2
+```
+
```
rapphil, line 75: I'm assuming that `provider` will be the only key that will be common among all the providers. all the rest of the keys will be provider dependant?

Maybe to solve this doubt, add the general format of a secret provider map:

e.g.: 
```
                   password:
                          provider:  <type of the provider>
                          <property1>: <property2>
```

My suggestion is to make explicit that `<type of the provider>` will be one of the supported list of providers and that all the rest of the keys is provider dependant.

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,136 @@
...
+  * https://github.com/prometheus/prometheus/issues/5795
+
+* **Other docs or links:**
+  * [Prometheus Remote Secrets Doc](https://docs.google.com/document/d/1EqHd2EwQxf9SYD8-gl3sgkwaU6A10GhiN7aw-2kx7NU/edit?tab=t.0)
+    * Previous proposal by @TheSpiritXIII
+  * https://stackoverflow.com/questions/43609144/can-prometheus-store-basic-auth-passwords-in-any-format-other-then-plain-text
+  * https://groups.google.com/g/prometheus-users/c/yWLE9qoG5GU/m/ke8ewxjIAQAJ
+  
+
+> TL;DR: This document proposes adding a new way for Prometheus to discover and use secrets from various secret providers, similar to how service discovery works. It introduces a new configuration section where users can specify different secret providers and their configurations. It also defines interfaces and methods for secret providers to implement, allowing for flexibility in how secrets are fetched and managed.
```
rapphil, line 23: Do we need to update the summary here. We will not have a separated section anymore.

hsmatulisgoogle: Done! Thanks

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Done!

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Done, thanks!

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,147 @@
...
+Secret providers might require secrets to be configured themselves. We will allow secrets to be passed in to secret providers.
+
+```
+...
+        password: !secret id=bootstrapped secret_id=pass1
+          provider: bootstrapped
+          secret_id: pass1
+          auth_token:
+            provider: kubernetes
+            secret_id: auth_token
```
rapphil, line 136:
```suggestion
...
        password:
          provider: bootstrapped
          secret_id: pass1
          auth_token:
            provider: kubernetes
            secret_id: auth_token
```
I think you forgot to remove the tag here.

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+However, an initial implementation might only allow inline secrets for secret providers. This might limit the usefulness of certain providers that require sensitive information for their own configuration.
+
+### Where will code live
+
+Both the alertmanager and prometheus repos will be able to use secret providers. The code will eventually live in a separete repository specifically created for it.
```
rapphil, line 143:
```suggestion
Both the Alertmanager and Prometheus repos will be able to use secret providers. The code will eventually live in a separete repository specifically created for it.
```

hsmatulisgoogle: Fixed thanks!

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Fixed as well thank you

rapphil (approved):

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,147 @@
...
+        password: my_important_secret
+```
+
+### Error handling
+
+We can classify all errors stemming from secret provider failures into 2 cases.
+
+The first case is that we have not gotten any secret value since startup. In this case we should not initialize any component that mentions this faulty secret, log this failure and schedule a retry to get this secret value and initialize the component.
+
+The second case is that we already have a secret value, but refreshing it has resulted in an error. In this case we should keep the component that uses this secret running with the potentially stale secret, schedule a retry and
```
rajagopalanand, line 100: Did you mean to add something more here?

rajagopalanand (commented):
LGTM

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,147 @@
...
+          auth_token:
+            provider: kubernetes
+            secret_id: auth_token
+```
+
+However, an initial implementation might only allow inline secrets for secret providers. This might limit the usefulness of certain providers that require sensitive information for their own configuration.
+
+### Where will code live
+
+Both the Alertmanager and Prometheus repos will be able to use secret providers. The code will eventually live in a separete repository specifically created for it.
```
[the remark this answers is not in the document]
bwplotka: That would be one option, but only if it's a separate Go module. I think it's fine to skip this detail in the proposal, and just decide later.

proposals/2025-02-24_secret_providers.md
bwplotka: +1

bwplotka (approved):
LGTM generally! Just last nit noted by @rajagopalanand and good to go IMO!

Approving from my side, but I will ask around for a second Prometheus maintainer to have another look

rajagopalanand (approved):

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,147 @@
...
+# HELP prometheus_remote_secret_state Describes the current state of a remotely fetched secret.
+# TYPE prometheus_remote_secret_state gauge
+prometheus_remote_secret_state{provider="kubernetes", secret_id="pass1", state="none"} 0
+prometheus_remote_secret_state{provider="kubernetes", secret_id="auth_token", state="stale"} 1
+prometheus_remote_secret_state{id="myk8secrets", secret_id="pass2", state="error"} 2
+```
+
+### Nested secrets
+
+Secret providers might require secrets to be configured themselves. We will allow secrets to be passed in to secret providers.
```
machine424, line 127: IIUC, in the Prometheus-on-Kubernetes setup, if the secret provider requires auth (which seems likely, I wouldn't see the point) then Prometheus will need to store the credentials for talking to the secret provider using the same mechanism (Kubernetes Secrets) that it currently uses to store the credentials it retrieves, thus prone to the same "vulnerabilities". Am I missing something?

bwplotka: Nice, thanks for double checking this.

This proposal is not about vulnerabilities (although you will have more choices on what providers to choose, based on security levels), but to all custom providers like we allow custom service discoveries.

The main outcome with the "recursive/nested" sort of approach here is that secret provider secret can't use the same secret provider and cycling connections are not allowed. Eventually secret provider has to use file (recommended) or inline (with environment variable eval for example). 

To avoid tricky implementation we could immediately limit the design to exactly that - provider secret can only use inline or file based, but it might be cleaner to not limit the choice here (e.g. there might be other providers that do not require secret)

machine424: Thanks for the details.

I was confused by `security` and `vulnerabilities` in `The motivation behind this design document is to enhance the security and flexibility of secret management in Prometheus.[...] can lead to security vulnerabilities. [...]`, I'd reformulate that as the proposal is mainly about adding an interface to how to fetch secrets and is not directly enhancing security per se, because the same storing mechanisms will continue to be used anyway to avoid chicken and egg situation (inline and file will continue to be used for secret providers secrets as you mentioned...)

In order to avoid falling back into the same challenges we have with the SD mechanisms, maybe we should be more explicit about ownership/maintenance, conditions for adding a new provider etc.
Also, did you consider only having a generic spec like `http_sd_config` and delegate the adapters addition/maintenance to the community/users?

hsmatulisgoogle: Thanks for the detailed comments! 

To clarify a little bit about what is meant when referring to the security issues, the main use case for  our team in proposing this related to the kubernetes environment where secrets stored on disk can be more easily exposed than those on memory

> because the same storing mechanisms will continue to be used anyway to avoid chicken and egg situation (inline and file will continue to be used for secret providers secrets as you mentioned...)

This chicken and egg issue wouldn't apply to kubernetes specifically since the container prometheus is running on will naturally have access to the kubernetes secret API as configured by kubernetes. This is also common for some cloud providers, where there is an identity associated with the machine you are running on that will naturally give you access to the related secret APIs (assuming you have configured it with the correct access rights)

> Also, did you consider only having a generic spec like http_sd_config and delegate the adapters addition/maintenance to the community/users?

I feel like only have a generic spec would be too restrictive and slow down adoption, as I imagine it requiring running sidecars in certain enviroments. Additionally designing and maintaining such a spec can be quite burdensome as there are many secret providers each with their own nuances. However in theory a generic spec could also be added as a secret provider

> In order to avoid falling back into the same challenges we have with the SD mechanisms, maybe we should be more explicit about ownership/maintenance, conditions for adding a new provider etc.

Yes, I understand the concern there about avoiding the issues with ownership/maintenance. I believe we could take a multi pronged approach like the following:

1. For common secret providers, support them directly
2. For less common secret providers either:
  i. Use a generic spec requiring a sidecar as mentioned above
  ii. Create a plugin system requiring custom compilation of prometheus

I am not sure what processes we could use to draw the line between `less common` and `common` secret providers. Do you have any ideas about an approach like this one?

bwplotka: Yea, we have the same challenges as with the SD mechanism. I don't think it's a blocker to this though, as the the same solution and path for SD could be used for SP (secret providers). I think it's fine to delegate this and solve the problem together with SD, not trying to come up with some unique solutions for SP that won't fit to SD later on.

However, in some way we are trying something new because we want to potentially maintain those in different repo (e.g. so AM don't need to import Prometheus), but it's TBC, don't need to be part of this proposal decision perhaps.

> For less common secret providers either:
> i. Use a generic spec requiring a sidecar as mentioned above

Good idea, but we tried that in the past and it's a bit too troublesome for users in practice, too much overhead and not as powerful as OOTB experience.

> ii. Create a plugin system requiring custom compilation of prometheus

yea, this is what we do with SD now and we should do the same in SP: https://github.com/prometheus/prometheus/blob/main/plugins.yml

proposals/2025-02-24_secret_providers.md
hsmatulisgoogle: Fixed thanks!

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,147 @@
...
+
+
+### Configuration
+
+Wherever a `<secret>` type is present in the configuration files, we will allow users to specify a map specifying how to fetch from a secret provider. The generic format would be
+
+```
+secret_field:
+  provider:  <type of the provider>
+  <property1>: <value1>
```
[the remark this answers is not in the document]
ArthurSens: I agree with David here, different secret providers might need different information to fetch a field. 

```go
type SecretField struct {
  Provider any # I know it's not really any
  property1 any
  property2 any
}
```
it's forcing all providers to have the same configuration options.

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,159 @@
...
+When specifying secrets inline in the config file, a string can be passed in as usual for backwards compatibility.
+```
+        password: my_important_secret
+```
+
+### Error handling
+
+We can classify all errors stemming from secret provider failures into 2 cases.
+
+The first case is that we have not gotten any secret value since startup. In this case we should not initialize any component that mentions this faulty secret, log this failure and schedule a retry to get this secret value and initialize the component.
```
ArthurSens, line 98: In case we never get a response, should we fail Prometheus's start-up or start with a missing secret?

Or even if we're in the process of retrying, should we delay Prometheus's start-up or start things anyway?

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,159 @@
...
+The time since the last successful secret fetch
+
+```
+prometheus_remote_secret_last_successful_fetch_seconds{provider="kubernetes", secret_id="pass1"} 15
+```
+
+A state enum describing in which error condition the secret is in:
+* error: there has been no successful request and no secret has been retrieved
+* stale: a request has succeeded but the latest request failed
+* none: the last request was succesful
```
ArthurSens, line 127: what do you think about "success" to represent success? Was there any particular reason to call it none?

ArthurSens (commented):
Looks pretty solid :) I just had some small questions

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
bwplotka: Yea, that's why @TheSpiritXIII (initial author) opted initially for

```
  password:
  password_file:
  password_ref: <ref to actual provider e.g. kuberentes with fields_for_kubernetes>
```

I think what @dgl proposes  and challenges @ArthurSens mention makes sense - it allows at least this consistency with e.g. scrape config discovery which follows the same pattern.

I DO however vote for at least attempting in practice to implement the variance of 


```
  password: <just "string" OR full complex provider(s) strcut>
```

Which avoids us to have this odd `_ref` or `_complex` duplicates everywhere.

So... how to find out what's possible? Can we right now implement a 3 Go files that show yaml unmarshalling and code for 3 alternatives we discuss here:

* This proposal now 
* @dgl proposal with `provider: kubernetes; kubernetes: <fields for kubernetes>` AND that variance for "inline string" OR complex providers secret struct"
* The most naive way so `password: "inline"; password_file: "..."; password_complex: <what we discuss here>`?

We could then compare easily all. WDYT? @hsmatulisgoogle ?

Essentially what will help us find the best solution and consensus here (:

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,159 @@
...
+  <propertyN>: <valueN>
+```
+
+For example when specifying a password fetched from the kubernetes provider with an id of `pass2` in namespace `ns1` for the HTTP passsword field it would look like this:
+
+```
+        password:
+          provider: kubernetes
+          namespace: ns1
+          secret_id: pass2
```
bwplotka, line 84: What's secret ID? Maybe better to make it obvious e.g.		

```suggestion
          provider: kubernetes
          namespace: <ns>
	  name: <secret name>
          key: <data's key for secret name>
```

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
bwplotka: I couldn't stop myself from experimenting here...

Here is the showdown of (1) question https://gist.github.com/bwplotka/baf5344837d98719dd126ed6bd143015 (the same in [playground](https://go.dev/play/p/qMXOvSwZ_v1)).

Check the "A" code (what this proposal proposes now). I think it does not look too magic or bad. Totally doable and trivial. I also like option "C" (somewhat what @dgl and @ArthurSens propose). A tiny bit more explicit, more consistent with scrape job + SD... we could do that too!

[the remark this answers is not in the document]
bwplotka: And to answer (2) added `secret2_test` to gist https://gist.github.com/bwplotka/baf5344837d98719dd126ed6bd143015 ([playground](https://go.dev/play/p/d1xZIDE5yby))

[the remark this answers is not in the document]
bwplotka: To sum up; learnings for me is to NOT worry about yaml unmarshal implementation. Generally everything discussed here is similarly safe and relatively trivial to implement. **Let's focus on what we want user to be able to do**

> 1. Dynamic provider fields VS per provider sections.

Dynamic is trivial to implement:

```go
func (s *SecretA) UnmarshalYAML(_ context.Context, unmarshalFn func(any) error) error {
	provider := struct {
		Provider string `yaml:"provider"`
	}{}
	if err := unmarshalFn(&provider); err != nil {
		return err
	}

	switch provider.Provider {
	case "kubernetes":
		s.Provider = provider.Provider
		return unmarshalFn(&s.Kubernetes)
	case "file":
		s.Provider = provider.Provider
		return unmarshalFn(&s.File)
	default:
		return fmt.Errorf("unknown provider %q", provider.Provider)
	}
}
```

...but I like the consistency of @dgl and @ArthurSens case:

```go
cExample = `
password:
  kubernetes:
    namespace: "<ns>"
    name: "<secret name>"
    key: "<data's key for secret name>"
`
	cExampleFile = `
password:
  file:
    path: "<path to secret file>"
`
```

> 2. Field with the dynamic "inline OR providers" type VS separate fields for inline, file and providers

Dynamic inline is trivial to implement as well, I vote for doing this.

```go
func (s *SecretA) UnmarshalYAML(_ context.Context, unmarshalFn func(any) error) error {
	// Try inlined form first, for compatibility and ease of use.
	var inlinedForm string
	if err := unmarshalFn(&inlinedForm); err == nil {
		s.Inline = InlineSP{Password: inlinedForm}
		return nil
	}
	// Fallback to complex struct.
	// NOTE: "plain" casting is needed to avoid recursive call to UnmarshalYAML.
	// This is what the current Prometheus code is doing too.
	type plain SecretA
	return unmarshalFn((*plain)(s))
}
```

bwplotka (commented):
Amazing feedback and discussion in https://github.com/prometheus/proposals/pull/47/files#r2041335383

Should we update the proposal with the mentioned alternatives considered?

What do you think is the reasonable solution from this discussion? I see the potential for changing proposal slightly to support the following (for consistency with SD):

```yaml
password: "<inlined secret>"
```

```yaml
password:
  kubernetes:
    namespace: "<ns>"
    name: "<secret name>"
    key: "<data's key for secret name>"
```

```yaml
password:
  file:
    path: "<path to secret file>"
```

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
dgl: 👍 

The only thing the experiment isn't addressing is that for example `SecretA` brings in the Kubernetes struct, which in a real implementation could end up bringing in the Kubernetes client libraries. Using some kind of dynamic registration like discovery does could potentially be used to break this dep.

However definitely agree with:
> Let's focus on what we want user to be able to do

I think the style you've outlined is good and this is enough for the proposal.

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Why
+
+The motivation behind this design document is to enhance the security and flexibility of secret management in Prometheus. Currently, Prometheus only supports reading secrets from the filesystem or directly from the configuration file, which can lead to security vulnerabilities and limitations when working with certain service providers.
+
+This proposal introduces secret discovery, similar to service discovery, where different secret providers can contribute code to read secrets from their respective APIs. This would allow for more secure and dynamic secret retrieval, eliminating the need to store secrets in the filesystem and reducing the potential for unauthorized access.
+
+### Pitfalls of the current solution
+
+Storing secrets in the filesystem poses risks, especially in environments like Kubernetes, where any pod on a node can access files mounted on that node. This could expose secrets to attackers. Additionally, configuring secrets through the filesystem often requires extra setup steps in some environments, which can be cumbersome for users.
```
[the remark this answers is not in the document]
machine424: Yes, related/same observation I had here https://github.com/prometheus/proposals/pull/47#discussion_r2034755085, even with this proposal we'll continue passing (meta) secrets inline or via file, I don't think the proposal should claim that the changes will improve on security.

proposals/2025-02-24_secret_providers.md
machine424: Thanks for the details.

> This chicken and egg issue wouldn't apply to kubernetes specifically since the container prometheus is running on will naturally have access to the kubernetes secret API as configured by kubernetes.

And AFAIK, those credentials live in the container's FS...
As long as we (should) use the inline and file for the secret providers secrets themselves, I still don't think this proposal should mention "security enhancement"; it indeed will help avoid storing (or having to provide) final secrets as inline or files, but users will still need to provide the meta secrets as inline or files.

Yes, I don't think the "ownership/maintenance" is a blocker, we can also add that later, I just wanted to draw attention to that. Maybe we can mention this in the proposal, that we'll work on that alongside SD.

 I like the idea of build time plugins as for SD in https://github.com/prometheus/prometheus/blob/main/plugins.yml
not sure how that works, but if we could have that also address https://github.com/prometheus/proposals/pull/47#discussion_r2046883686 to control/limit the extra dependencies, that'd be great.

A last one, I see a lot of focus on kubernetes secret provider (that would be a good candidate for OOTB/enabled by default plugin), do you know what the other providers could be? could we list some of them, it'd help with thinking about the API, with implementation.

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
bwplotka: > The only thing the experiment isn't addressing is that for example SecretA brings in the Kubernetes struct, which in a real implementation could end up bringing in the Kubernetes client libraries. Using some kind of dynamic registration like discovery does could potentially be used to break this dep.

Yea, true, then we might need some dynamic "any" struct anyway for that reason, but I wanted to focus on a user side first.

rajagopalanand:
I vote for changing the proposal to be consistent with SD (What @bwplotka proposed at the end)

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
hsmatulisgoogle: I switched the design doc to use the more consistent struct style, as mentioned i the main thread

proposals/2025-02-24_secret_providers.md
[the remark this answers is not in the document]
hsmatulisgoogle: I removed the claims on security, I agree there doesn't seem to be a substantial benefit to security. Thanks for pointing it out!

proposals/2025-02-24_secret_providers.md
hsmatulis: I added a paragraph clarifying the design here, thanks!!

proposals/2025-02-24_secret_providers.md
hsmatulis: That is a much better choice, thanks!

proposals/2025-02-24_secret_providers.md
hsmatulis: Thanks! accepted it

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,253 @@
...
+### Secret rotation
+
+When a secret is rotated, it is possible for this to happen out of sync. For  example, one of the following could happen with secrets used for target scraping:
+
+1. The scrape target's secret updated but the currently stored secret hasn't updated yet.
+2. The secret provider's secret changes but the scrape target hasn't updated to the new secret yet.
+
+Therefore to help alleviate 1. we can ignore caching and query the secret provider again if permission errors are reported by the component.
+
+To alleviate 2. we can store the previous secret value before rotation for a period of time and use it as a fallback in case of permission errors.
```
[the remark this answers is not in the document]
hsmatulis: Added the section, thanks!!

hsmatulis:
I changed the config style to be more consistent with SD in the proposal. Thanks everyone for all the feedback, and feel free to reach out on the prometheus slack if you have any feedback or ideas. I am working on an implementation for the proposal:)

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,277 @@
+## Remote Secrets (Secret Providers)
+
+* **Owners:**
+  * Henrique Matulis (@hsmatulisgoogle)
```
dgl, line 4: nit: username doesn't match (anymore? I think it was changed)

dgl (approved):

machine424 (approved):
lgtm, thanks!

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,277 @@
...
+secret_namespace = "my-application-namespace"
+secret_name = "my-database-secret"
+
+// Step 3: Retrieve the secret object from the Kubernetes API
+secret_object = get_kubernetes_secret(secret_namespace, secret_name, kubernetes_client)
+```
+
+## Action Plan
+
+* [ ] Create action plan after doc is stable!
```
ArthurSens, line 277: Do you want to work on this before or after merging? :)

ArthurSens (approved):
LGTM, one last question :)

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,277 @@
...
+
+For example when specifying a password fetched from the kubernetes provider with an id of `pass2` in namespace `ns1` for the HTTP passsword field it would look like this:
+
+```
+        password:
+          kubernetes:
+            namespace: <ns>
+            name: <secret name>
+            key: <data's key for secret name>
+```
```
rapphil, line 86: the example is not matching the statement. 

Did you mean:
```suggestion
For example when specifying a password fetched from the kubernetes provider with an id of `pass2` in namespace `ns1` for the HTTP passsword field it would look like this:

```
        password:
          kubernetes:
            namespace: ns1
            name: pass2
            key: http_password
```
```

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,277 @@
...
+      password:
+        vault: # Example provider
+          address: "https://vault.example.com"
+          path: "secret/data/prometheus/example_job"
+          key: "password"
+          refresh_interval: 30m # Override default for this secret
+    # ... other config ...
+```
+* **Caching Behavior on Failure:** If a scheduled background refresh attempt fails (e.g., due to network issues, temporary provider unavailability, invalid credentials after rotation), Prometheus will continue to use the last successfully fetched secret value. This ensures components relying on the secret can continue operating with the last known good credential, preventing outages due to transient refresh problems. Failed refresh attempts will be logged, and the `prometheus_remote_secret_state` metric will reflect the 'stale' state.
+* **Permission Errors:** If a component using a cached secret reports a specific permission or authentication error, this might indicate the cached secret is no longer valid. Hence, an immediate refresh attempt for that specific secret is triggered, bypassing the regular refresh interval.
```
rapphil, line 126: I'm a bit concerned about the implications of this statement. Isn't this requirement for secrets providers leaking into error handling strategy of all components that use a secrets provider? 

We only know that a credential is expired/revoked after we try to use it. My point is that all components that use secrets provider will need to implement error handling in a specific way or will the secret provider provide some functionality that components should use?

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,277 @@
...
+
+Metrics should be present to help users identify the errors mentioned above (in addition to logs). Therefore the following metrics should be reported per secret present in the config file:
+
+The time since the last successful secret fetch
+
+```
+prometheus_remote_secret_last_successful_fetch_seconds{provider="kubernetes", secret_id="pass1"} 15
+```
+
+A state enum describing in which error condition the secret is in:
```
rapphil, line 151: In addition to the state, maybe it would be useful to also expose some counters to show events. Why? because that allow us to identify precisely what is happening across multiple scrapes, network failures etc. Something like:

```
prometheus_remote_secret_fetch_failures_total
prometheus_remote_secret_fetch_success_total
```

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,277 @@
...
+          key: "password"
+          refresh_interval: 30m # Override default for this secret
+    # ... other config ...
+```
+* **Caching Behavior on Failure:** If a scheduled background refresh attempt fails (e.g., due to network issues, temporary provider unavailability, invalid credentials after rotation), Prometheus will continue to use the last successfully fetched secret value. This ensures components relying on the secret can continue operating with the last known good credential, preventing outages due to transient refresh problems. Failed refresh attempts will be logged, and the `prometheus_remote_secret_state` metric will reflect the 'stale' state.
+* **Permission Errors:** If a component using a cached secret reports a specific permission or authentication error, this might indicate the cached secret is no longer valid. Hence, an immediate refresh attempt for that specific secret is triggered, bypassing the regular refresh interval.
+* **Not Querying Per-Scrape:** It is explicitly **not** the default behavior to query the secret provider on every use.
+
+
+### Secret rotation
```
rapphil, line 130: Is there any example of backend that would benefit from the behaviours defined in this section.

Any example of backend that does not support secret overlapping?

proposals/2025-02-24_secret_providers.md
rajagopalanand: If we take PagerDuty or Slack notifier as an example, is the expectation those integrations/receiver implementations should interpret AuthN errors returned and somehow notify the secret provider to refresh before retrying?

proposals/2025-02-24_secret_providers.md
HenriqueMatulis: sure, added thanks:)

proposals/2025-02-24_secret_providers.md
hsmatulis: Fixed, thanks!

proposals/2025-02-24_secret_providers.md
hsmatulis: sure, added thanks:)

proposals/2025-02-24_secret_providers.md
hsmatulis: Ah yes, thank you!!

proposals/2025-02-24_secret_providers.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Error handling
+
+We can classify all errors stemming from secret provider failures into 2 cases.
+
+The first case is that we have not gotten any secret value since startup. In this case we should not initialize any component that mentions this faulty secret, log this failure and schedule a retry to get this secret value and initialize the component.
+
+The second case is that we already have a secret value, but refreshing it has resulted in an error. In this case we should keep the component that uses this secret running with the potentially stale secret, and schedule a retry.
+
+If we are starting up prometheus and do not get any secret values, startup will continue and prometheus will continue to run and retry finding secrets. If there are components that are fully specified, they will run during this time. The idea is that it is better to send partial metrics than no metrics, and the emitted metrics for secrets can alert users if there is a problem with their service providers.
```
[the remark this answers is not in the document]
hsmatulis: Fixed, thanks!

proposals/2025-02-24_secret_providers.md
hsmatulis: > We only know that a credential is expired/revoked after we try to use it. My point is that all components that use secrets provider will need to implement error handling in a specific way or will the secret provider provide some functionality that components should use?

Good question! The idea would be to signal the background secret provider process manually whenever a refresh is needed, that way we have more flexibility in the implementations. I added this discussion to the doc.

> If we take PagerDuty or Slack notifier as an example, is the expectation those integrations/receiver implementations should interpret AuthN errors returned and somehow notify the secret provider to refresh before retrying?

Yes, that is the gist of it. Currently the AlertManager already has [a centralized retry loop that we can hook into](https://github.com/prometheus/alertmanager/blob/e060127980825a83511f651d1eb48660fc74efd6/notify/notify.go#L791-L899) that classifies each secret type as retryiable. We could change that to an enum and add a type of authenication error, or choose to always retry. The specific way to implement it would be in the implementation PR though

Thanks to you both!!!

proposals/2025-02-24_secret_providers.md
hsmatulis: Agreed & added. Thanks!

proposals/2025-02-24_secret_providers.md
hsmatulis: > Is there any example of backend that would benefit from the behaviours defined in this section.

This section is less about the secret provider (Vault, AWS) and more about the target system's authentication behaviour during rotation. This can happen if there is any sort of caching during the auth process, or even just a race condition

Specific example:

The target application doesn't query the secret provider directly. Instead, a sidecar pulls the secret (e.g., from AWS Secrets Manager) every 5 minutes and updates a local file/cache the target reads.
Prometheus gets `new_pass` immediately after rotation from AWS.
If the sidecar hasn't refreshed its local cache yet, the target does not knows of `new_pass` so it only accepts `old_pass`.

rapphil (approved):

ArthurSens (approved):
Thanks for workin on the proposal, I'm merging as soon as you tell me you're satisfied :)

hsmatulis:
I think I am ready to merge it, thanks everyone!

sysadmind (approved):

PR #48, opened 2025-03-27
Proposal: OTEL delta temporality support

fionaliao:
A proposal for https://github.com/prometheus/prometheus/issues/12763


fionaliao:
Updates:
* Simplified proposal - moved CT-per-sample to possible future extension instead of embedding within proposal
* Changed proposal to have a new `__temporality__` label instead of extending `__type__` - probably better to keep metric type concept distinct from metric temporality. This also aligns with how OTEL models it.
* Updated remote-write section - delta ingestion will actually be fully supported via remote write (since CT-per-sample is moved out of main proposal for now)
* Moved temporary `delta_rate()` and `delta_increase()` functions suggestion to discarded alternative - not sure this is actually necessary if we have feature flag for temporality-aware functions anyway
* Fleshed out implementation plan

proposals/2025-03-25_otel-delta-temporality-support.md
```diff
@@ -0,0 +1,361 @@
...
+
+`irate()` will also be extended to support delta metrics.
+
+Having functions transparently handle the temporality simplifies the user experience - users do not need to know the temporality of a series for querying, and means queries don't need to be rewriten wehn migrating between cumulative and delta metrics.
+
+`resets()` does not apply to delta metrics, however, so will return no results plus a warning in this case.
+
+#### rate() calculation
+
+In general: `sum of second to last sample values / (last sample ts - first sample ts)) * range`. We skip the value of the first sample as we do not know its interval.
```
enisoc, line 152: > We skip the value of the first sample as we do not know its interval.

Perhaps we could get utility out of the first sample's value by guessing that each sample's interval is equal to the average time between samples in the window. One motivation for this is a case that we see often with non-sparse deltas produced by an intermediate processor.

Suppose the actual instrumented app is sending delta samples at a regular 60s interval. We'll assume for simplicity that these deltas are always greater than zero. Then there is an intermediate processor that's configured to accumulate data and flush every 30s. To avoid spareness, it's configured to flush a 0 value if nothing was seen.

The data stream will look like this, with a sample every 30s:

```
5 0 2 0 10 0 8 0
```

Note that every other value is 0 because of the mismatch between the flush intervals of the instrumented app and the intermediate processor.

If we then do a `rate(...[1m])` on this timeseries, with the current proposal, we might end up with the 1m windows broken up like this:

```
5 0 | 2 0 | 10 0 | 8 0
```

If we can't make use of the value from the first sample in each window, we will end up computing a rate of `0` for all of these windows. If instead we guess that each sample represents the delta for a 30s interval, because that's the average distance between the two datapoints in the window, then we will compute the correct rates. Of course it was only a guess and you could contrive a scenario that would fool the guess, but the idea would be to assume that the kind of scenario described here is more likely than those weird ones.

proposals/2025-03-25_otel-delta-temporality-support.md
```diff
@@ -0,0 +1,361 @@
...
+
+We could just not try and predict the start/end of the series and assume the series continues to extend to beyond the samples in the range. However, not predicting the start and end of the series could inflate the rate/increase value, which can be especially problematic during rollouts when old series are replaced by new ones.
+
+Assuming `rate()` only has information about the sample within the range, guessing the start and end of series is probably the least worst option - this will at least work in delta cases where the samples are continuously ingested. To predict if a series has started ended in the range, check if the timestamp of the last sample are within 1.1x of an interval between their respective boundaries (aligns with the cumulative check for start/end of a series).
+
+To calculate the interval, use the average spacing between samples.
+
+Downsides:
+
+* This will not work if there is only a single sample in the range, which is more likely with delta metrics (due to sparseness, or being used in short-lived jobs).
```
enisoc, line 166: For cumulative samples, it makes sense that with a single sample in the window you can't guess anything at all about how much increase happened in that window. With a single delta sample, even if we don't know the start time, we should be able to make a better guess than "no increase happened".

For example, we could guess that the interval is equal to the window size -- in other words return the single delta value as is with no extrapolation. The argument would be that you picked an arbitrary window of size T and found 1 sample, so the best guess for the frequency of the samples is 1/T. This seems like it would be more useful on average than returning no value in the case of a single sample.

proposals/2025-03-25_otel-delta-temporality-support.md
fionaliao: A cumulative series could generate the same values (with the zeroes being where the series resets). And in that case rate() would return no results. So though this doesn't accurately capture the rate, the behaviour would be consistent for cumulative and delta metrics.

However, the first point in a window is more meaningful in the delta case - you know it's a delta from the preceeding sample while in the cumulative case you have to look outside the window to get the same information, so maybe we should do better because of that. That example is leaning me more towards "just do sum_over_time() / range for delta rate()" - in this case that would probably give more useful information. Or at least do that before CT-per-sample available, at which point we'd have more accurate interval data.

proposals/2025-03-25_otel-delta-temporality-support.md
fionaliao: My concern is with mixing extrapolation and non-extrapolation logic because that might end up surprising users. 

if we do decide to generally extrapolate to fill the whole window, but have this special case for a single datapoint, someone might rely on the non-extrapolation behaviour and get surprised when there are two points and it changes .

fionaliao:
Next steps for this proposal:
- Wait for type and unit metadata proposal to be finalised, which might result in updates to how exactly the temporality label will be implemented
- Get the primitive otel delta support PR merged, hopefully having that out will help get some feedback on querying should be done
- Write some code to experiment with delta rate implementations, see what edge cases there are for each option

fionaliao:
> Write some code to experiment with delta rate implementations

Started rough implementation for rate functions here:

https://github.com/prometheus/prometheus/compare/fionaliao/basic-delta-support...fionaliao/delta-rate

Including some example queries: https://github.com/prometheus/prometheus/blob/4c72cba2e76ac55c77c46af7b2b9348e8cf67b59/promql/promqltest/testdata/delta.test

beorn7 (commented):
Thanks for this design doc.

I realize that my comments are a bit all over the place, and often they discuss things that are already taken into account in this document, just maybe not in the order or emphasis I would prefer.

An attempt to summarize my thoughts and concerns:

I wholeheartedly agree with "Milestone 1". However, I do think we should let the experiences made from it inform our further steps. The design doc explains most of the possible approaches quite well, but it essentially already proposes a preferred solution along the following lines:
1. Introduce a temporality label.
2. Make rate/increase/irate behave differently depending on that label.
3. Embrace an extrapolation approach in that rate/increase calculation.

I have concerns about each of these points. I wouldn't go as far to say that they are prohibitive, but I would have trouble approving a design doc that frames them as the preferred way to go forward, while the alternatives that I find more likely to be viable are already "demoted" to "alternatives that we have dismissed for now".

My concerns summarized:
1. I would only introduce a temporality label once we have established we need one. I would go for "treat deltas as gauges" until we hit a wall where we clearly see that this is not enough. In the form of the outcome of recording rules, Prometheus had "delta samples" from the beginning, and never did we consider marking them as such.
2. I have a very bad feeling about "overloading" and have certain functions behave fundamentally different depending on the type of the argument (and that even more so as we are just establishing this strong typing of metrics as we go). (We kind-of do that already for some histogram functions, but there the difference in type is firmly established, plus it's not really fundamentally different what we are doing, we are doing "the same" on different representations of histograms (classic vs. native), plus we will just get rid of the "classic" part eventually.) Additionally, I don't think it makes sense to claim that we are calculating an "increase" based on something that is already an increase (a delta sample). The "rate'ing" is then just the normalization step, which is just one part of the "actual" rate calculation. Even though it might be called that way in other metrics system, I don't think that should inform Prometheus naming. I do understand the migration point, but I see it more as a "lure" into something that looks convenient at first glance but has the potential of causing trouble precisely because it is implicit (or "automagic"). What might convince me would be a handling of ranges that contain "mixed samples" (both cumulative and delta samples) because that would actually allow a seamless migration, but that would open up a whole different can of worms.
3. Extrapolation caused a whole lot of confusion and controversy for the existing rate conversion. I believe that it was necessary, but I see a different trade-off for delta samples. Given that we have active work on non-extrapolation (`anchored` in the PoC) and "a different kind of interpolation" (`smoothed` in the PoC) for rate calculation, we should hold back introducing a questionable extrapolation mechanism in delta handling. With the tracking of CT (aka StartTimeUnixNano), we are well set up to do something like `smoothed` for deltas (which is something to be fleshed out maybe here or maybe in a separate design doc), and in many cases, the naive non-extrapolation approach might just be the best for deltas. (An "aligned" rule evaluation feature might be easier to implement and more helpful for the use case of aligned delta samples.)[^1]

To summarize the summary: I would pivot this design doc more as a list of alternatives we have to explore, and only state the first step as "already decided", namely to ingest the delta samples "as is", which puts us into a position to explore the alternatives in practice.

[^1]: if you feel that aligned rule evaluation and "smoothed" increase calculation from deltas should be included in this doc, I'm willing to flesh them out in more detail.

fionaliao:
@beorn7 Thanks for your comments :) I need more time to go through all of them, but as a start:

> To summarize the summary: I would pivot this design doc more as a list of alternatives we have to explore, and only state the first step as "already decided", namely to ingest the delta samples "as is", which puts us into a position to explore the alternatives in practice.

It makes sense to separate the first step. Not a strong opinion, but I was thinking of making a separate proposal PR for the first step, and then put this proposal back into draft to indicate it's still being figured out - that way we can have a merged PR for the first step, and there can be discussion on the possible future steps within this open PR. WDYT?

---

> I would only introduce a temporality label once we have established we need one. I would go for "treat deltas as gauges" until we hit a wall where we clearly see that this is not enough.

I agree with doing this delta as gauge approach first. I do think eventually we will want to treat deltas separately from gauges, but we should get more user feedback to confirm this is the case.

Chronosphere have already gained insights into on this, as they've implemented their own version of delta support and @enisoc wrote up [this document](https://docs.google.com/document/d/1L8jY5dK8-X3iEoljz2E2FZ9kV2AbCa77un3oHhariBc/edit?tab=t.0) and noted: "Users don't like having to think about the temporality of their metrics and learn to use different functions (e.g. increase vs. sum_over_time). They want to have one recommended way to query counters that just works regardless of temporality.".

One problem with treating deltas as gauges is that gauge means different things in Prometheus and OTEL - in Prometheus, it's just a value that can go up and down, while in OTEL it's the "last-sampled event for a given time window". While it technically makes sense to represent an OTEL delta counter as a Prometheus gauge, this could be a point of confusion for OTEL users who see their counter being mapped to a Prometheus gauge, rather than a Prometheus counter. There could also be uncertainty for the user on whether the metric was accidentally instrumented as a gauge or whether it was converted from a delta counter to a gauge.

Another problem is that the temporality of a metric might not be completely under the control of the user instrumenting the metric - it could change in the metric ingestion pipeline (e.g. with the [cumulativetodelta](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/cumulativetodeltaprocessor/README.md) or [deltatocumulative](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/deltatocumulativeprocessor) processors), so it can be hard to determine at query time what function to use. If we properly mark deltas as gauges - i.e. with the metric type gauge - and have warnings when using `rate()` on Prometheus gauges and `sum_over_time()` on Prometheus counters, this is alleviated. (However, alerts don't integrate with warnings so may end up being incorrect without detection).

> We kind-of do that already for some histogram functions, but there the difference in type is firmly established

How is type being firmly established in the native histogram case vs not being firmly established in the delta and cumulative case if there's a "temporality" label?

> Additionally, I don't think it makes sense to claim that we are calculating an "increase" based on something that is already an increase (a delta sample).

`increase()` could be considered the increase in the underlying thing being measured, which makes sense for applying `increase()` on a delta metric.

Also, deltas could also be seen as cumulative with resets between each sample. (On the other hand, as discussed, there are different characteristics of delta metrics so while they could be seen as cumulative or converted to cumulative that might not be the best representation)

> I do understand the migration point, but I see it more as a "lure" into something that looks convenient at first glance but has the potential of causing trouble precisely because it is implicit (or "automagic"). What might convince me would be a handling of ranges that contain "mixed samples" (both cumulative and delta samples) because that would actually allow a seamless migration, but that would open up a whole different can of worms.

As well as one-off migrations where you might just have to update queries once, a case which might cause more constant frustration is when there is a mix of sources with different temporalities. So a single series might have the same temporality over time, but different series have different temporalities. If you want a single query to combine the results and we didn't do function overloading, you'd need something like `rate(cumulative metrics only) + sum_over_time(delta metrics only)`. (Is this what you were referring to when you said mixed samples, or did you just mean the case where a single series had different temporality over time?)

proposals/2025-03-25_otel-delta-temporality-support.md
```diff
@@ -0,0 +1,364 @@
...
+
+#### Non-approximation
+
+While `rate()` and `increase()` will be consistent with cumulative case, there are downsides as mentioned above. Therefore there may be cases where approximating the rate/increase over the selected range is unwanted for delta metrics.
+
+Users may prefer "non-approximating" behaviour that just gives them the sum of the sample values within the range. This can be accomplished with `sum_over_time()`, though this does not always accurately give the increase within the requested range.
+
+`sum_over_time()` does not work for cumulative metrics, so a warning should be returned in this case. One downside is that this could make migrating from delta to cumulative metrics harder, since `sum_over_time()` queries would need to be rewritten, and users wanting to use `sum_over_time()` will need to know the temporality of their metrics.
+
+One possible solution would to have a function that does `sum_over_time()` for deltas and the cumulative equivalent too (this requires subtracting the latest sample before the start of the range with the last sample in the range). This is outside the scope of this design, however.
```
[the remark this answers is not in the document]
enisoc: > from the explanation so far, I would assume that a range with mixed samples will error out

At first glance, I think the current proposal of using a `__temporality__` label means that if you follow the best practice of taking rate/increase before aggregating, the migration should work fine. Each individual time series can't have mixed temporality samples because changing temporality changes the label value.

beorn7:
> I was thinking of making a separate proposal PR for the first step, and then put this proposal back into draft to indicate it's still being figured out

I don't think it would help with clarity to have multiple design docs. Personally, I don't believe a design doc has to make a call all the way through. I would think it's fine if a design doc says "We want to do this first step, and then we want to do research and want to make a call between options X, Y, and Z based on these following rationales."

About the general argument about "overloading" increase and rate for delta temporality: I think the arguments are already well made in the design doc. I'm just not sure we can make a call right now without practical experience. We can repeat and refine both sides of the argument, but I think it will be much easier and much more convincing once we have played with it.

> How is type being firmly established in the native histogram case vs not being firmly established in the delta and cumulative case if there's a "temporality" label?

First of all, that label does not exist yet. So currently, it is not established at all. Secondly, a histogram sample looks completely different from a float sample in the low-level TSDB data. There is no way the code can confuse one for the other. But a label is just a label. It could accidentally get removed, or added (or maybe even on purpose, "hard-casting" the metric type, if you want), so a relatively lightweight thing like a label will change how a function processes something that is just a float in the TSDB in either case.

> Is this what you were referring to when you said mixed samples, or did you just mean the case where a single series had different temporality over time?

I was thinking mostly about one and the same series that changes from cumulative to delta over time. (Your point about mixed vectors is also valid, but that would be solved by the proposed "overloaded" functions just fine.)

fionaliao:
@beorn7 I'll update this doc as you suggested (with the first step + laying out the options for future steps without committing to any), and incorporate you and @enisoc's comments

proposals/2025-03-25_otel-delta-temporality-support.md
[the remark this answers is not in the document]
beorn7: Yes, if we decided to make temporality port of the series identity, we would not get mixed series. But this isn't a complete solution yet:
- I would expect that we remove the `__temporality__` label after `rate`/`increase` (because the whole point is that the result doesn't need to know about its original temporality anymore), so we would get a collision after the operation. (We could mitigate that in a similar way as we did with "delayed name removal" – which is still an experimental feature, which we needed to include here to make it work.)
-  If we do not aggregate after `rate`/`increase`, we either get the collision from above (even with the "late removal" mitigation), or we get two separate time series for (naively) "the same" metric (if we somehow mark the resulting series from different temporality as different).

That's hinting a bit towards what I mean, that providing some "magic" can easily backfire and cause even more confusion. It's a relatively easy explanation to tell users that delta "counters" are a different beast and not really counters in the Prometheus sense and therefore have to be treated differently. However, if you provide the "magic" and tell users they don't have to worry about temporality, and then they get "vector cannot contain metrics with the same labelset", you have _way_ more to explain. Or they get two separate results for a metric they consider just one. Again _way_ more to explain.

Put in another way: If we provide magic, it really has to work reliably.

PromQL's traditional philosophy is more like Go's one: Avoid the magic, provide easy to understand primitives, and build thing out of them, even if it is sometimes a bit more verbose or clunky than with that fancy other language that has high level abstractions.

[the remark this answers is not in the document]
enisoc: For the sake of informing further brainstorming on the migration topic: In our custom TSDB that we run underneath the OSS PromQL engine, we went the route of having mixed-temporality samples within a single series. Every sample has optional metadata stored with it, which is only present if it's the first sample in the block, or if the value has changed since the previous sample. As we iterate over samples to return to the PromQL engine, we take note if the temporality changes and switch on/off the query-time delta-to-cumulative conversion as needed. This wouldn't solve the design constraints we've set for ourselves in this proposal (e.g. taking advantage of the extra information in delta samples rather than converting to cumulative), but I offer it as an example of how we've dealt with mixed temporality samples.

So in our case, we didn't encounter the vector collision when we migrated users, but of course it introduces other complexity (particularly at the TSDB level) that I had been hoping we could avoid with the `__temporality__` label idea. But as you pointed out, including temporality in the series identity comes with its own complexity.

Still, I'm fairly motivated to keep looking for ways to solve the migration case in a way that's transparent to people who write queries. In the OTel world, it's common for other roles in your org to make decisions that suddenly change the temporality of your metrics out from under you. OTel treats temporality as a low-level concern that isn't really visible at the level of instrumenting application code. It would be unfortunate if the only way to avoid disruption is to always write your queries in a complicated defensive form that somehow checks for both possibilities and applies different functions.

fionaliao:
@beorn7  Would you be open to having deltas ingested as gauges by default, with an option to ingest as counters with a `__temporality__="delta"` label? With documentation making it clear that all of this is experimental and could be removed. This won't include implementing any function overloading, it just adds the label so users can distinguish between delta counters and gauges if they want to.

I think to explore if it's worth persuing the `__temporality__` label, we should to offer it as an option and see how users interact with it.

beorn7:
Sounds good to me.

ArthurSens:
> @beorn7 Would you be open to having deltas ingested as gauges by default, with an option to ingest as counters with a `__temporality__="delta"` label? With documentation making it clear that all of this is experimental and could be removed. This won't include implementing any function overloading, it just adds the label so users can distinguish between delta counters and gauges if they want to.
> 
> I think to explore if it's worth persuing the `__temporality__` label, we should to offer it as an option and see how users interact with it.

Just to clarify, how would this optionality be granted to users? Is it yet a new feature flag? A config option?

ArthurSens:
Interesting, my idea was to eventually sunset the delta to cumulative conversion 🤔

I'm a bit concerned with adding new fields to config file because removing them would be a breaking change. Removing feature flags isn't as hard as removing config options

proposals/2025-03-25_otel-delta-temporality-support.md
```diff
@@ -0,0 +1,388 @@
...
+### Delta metric type
+
+It is useful to be able to distinguish between delta and cumulative metrics. This would allow users to understand out what the raw data represents and what functions are appropiate to use. Additionally, this could allow the query engine or UIs displaying Prometheus data apply different behaviour depending on the metric type to provide meaningful output. 
+
+There are two options we are considering for how to type deltas. We propose to add both of these options as feature flags for ingesting deltas.
+
+1. `--enable-feature=otlp-delta-as-gauge-ingestion`: Ingests OTLP deltas as gauges.
+
+2. `--enable-feature=otlp-native-delta-ingestion`: Ingests OTLP deltas as a new delta "type", using a new `__temporality__` label to explicitly mark metrics as delta.
+
```
alanprot, line 120: Should we consider a more genetic label to keep track of the original OTEL type (ex: `__otel_type__`) - and so we can this information to make other decisions in the future - similar to what we are proposing on the rate function here?

proposals/2025-03-25_otel-delta-temporality-support.md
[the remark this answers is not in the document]
fionaliao: Added the issue with migration / separate series for delta and cumulative. Note that if we follow the logic for type and unit metadata labels, they're dropped after `rate()` or `increase()` is applied and two series with the same labelset will be returned (with an info annotation about the query containing mixed types).

fionaliao:
This proposal is now ready for review again. The above comments have been address or incorporated into the proposal. 

Key callouts:
* **Especially interested in feedback for this:** Instead of adding `__type__="counter"` and `__temporality__="delta"` to OTEL deltas or provide an option to ingest as gauges and another option to ingest as counter +temporality, the proposal now is to just ingest them as gauges with additional `__otel_type__="sum"` and `__otel_temporality__` labels  (and the histogram equivalent), the reasoning being it's a less intrusive change to start with, with the option of going down the `__temporality__` route later. The downsides is that we sort of end up with two type systems in Prometheus (the OTEL one with the `__otel_...` labels and then the actual Prometheus one).
  * The previous updated proposal with the two ingestion options can be seen [here](https://github.com/prometheus/proposals/blob/1064e29cd43710c9483197e029d7592102581332/proposals/0048-otel_delta_temporality_support.md).
* Reduced the scope of the proposal in general - we will just add the appropriate metadata to the series being ingested and function warnings, with function overloading being a potential future extension.
* I'm less keen on now implementing a rate function for deltas that only looks at the samples within the range and then extrapolates to get the rate/increase for the entire range. This is because there's been a proposal accepted for adding (experimental) range selector modifiers (https://github.com/prometheus/proposals/blob/main/proposals/2025-04-04_extended-range-selectors-semantics.md). This has been proposed with the context of cumulative counters, however could probably be used in a similar way for deltas to estimate the increase/rate for the series within the selected query range, but since it can look beyond the boundary of the query range, it should get better results compared to just extrapolation. So the extrapolation option for deltas is probably not necessary.
* Added some discussion on monotonicity, which has been influential for me for going down the gauge route for now (allowing non-monotonic counters is a big change for the Prometheus model, but not allowing non-monotonic counters means statsd counters cannot be supported).

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,493 @@
...
+* The original delta information is lost upon conversion. If the resulting Prometheus gauge metric is converted back into an OTEL metric, it would be converted into a gauge rather than a delta metric. While there's no proven need for roundtrippable deltas, maintaining OTEL interoperability helps Prometheus be a good citizen in the OpenTelemetry ecosystem.
+
+It is useful to be able to identify delta metrics and distinguish them from cumulative metrics and gauges. This would allow users to understand what the raw data represents and what functions are appropriate to use. Additionally, this could allow the query engine or UIs displaying Prometheus data apply different behaviour depending on the metric type to provide meaningful output.
+
+Therefore it is important to maintain information about the OTEL metric properties. Alongside setting the type to `gauge` / `gaugehistogram`, the original OTEL metric properties will also be added as labels:
+
+* `__otel_type__="sum"` - this will allow the metric to be converted back to an OTEL sum (with delta temporality) rather than a gauge if the metrics as exported back to OTEL.
```
fionaliao, line 151: Wondering if we need the `__otel_type__` label since you can infer a metric is an OTEL sum if it has the `__temporality__` label (OTEL gauges don't have temporality).

Metrics ingested from other sources with these labels could then be accidentally converted to otel sums. That might be acceptable though, these labels would probably be added to indicate deltas

fionaliao:
Updated to remove the `__otel_` prefix when unnecessary post dev summit: https://docs.google.com/document/d/1uurQCi5iVufhYHGlBZ8mJMK_freDFKPG0iYBQqJ9fvA/edit?tab=t.0#heading=h.xp8na12byu2i

> Consensus: There are no objections to the proposal to ingest delta metrics directly with additional labels (sans the otel_ prefix) into the tsdb as gauges as an experimental feature.

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,491 @@
...
+* `__temporality__="delta"`
+* `__monotonicity__="true"/"false"` - as mentioned in [Monotonicity](#monotonicity), it is important to be able to ingest non-monotonic counters. Therefore this label is added to be able to distinguish between monotonic and non-monotonic cases.
```
ArthurSens, line 153: Did you mean `__monotonicity__="monotonic/non-monotonic"`? Or maybe `__monotonic__="true/false"`?

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,491 @@
...
+
+### Scraping
+
+No scraped metrics should have delta temporality as there is no additional benefit over cumulative in this case. To produce delta samples from scrapes, the application being scraped has to keep track of when a scrape is done and resetting the counter. If the scraped value fails to be written to storage, the application will not know about it and therefore cannot correctly calculate the delta for the next scrape.
+
+### Federation
+
+Federating delta series directly could be usable if there is a constant and known collection interval for the delta series, and the metrics are scraped at least as often as the collection interval. However, this is not the case for all deltas and the scrape interval cannot be enforced. 
+
+Therefore we will add a warning to the delta documentation explaining the issue with federating delta metrics, and provide a scrape config for ignoring deltas if the delta labels are set.
```
ArthurSens, line 185: Instead of giving a warning, could we update the federation endpoint to just not expose metrics with `__temporality__="delta"` and/or `__monotonicity__="non-monotonic"`?

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,493 @@
...
+
+Additionally, as mentioned before, it is common for deltas to have samples emitted at fixed time boundaries (i.e. are aligned). This means if the collection interval is known, and query ranges match the time boundaries, accurate results can be produced with `sum_over_time()`.
+
+#### Function warnings
+
+To help users use the correct functions, warnings will be added if the metric type/temporality does not match the types that should be used with the function. The warnings will be based on the `__type__` label, so will only work if `--enable-feature=type-and-unit-labels` is enabled. Because of this, while not necessary, it will be recommended to enabled the type and unit labels feature alongside the delta support feature.
```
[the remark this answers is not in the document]
ArthurSens: I think that's a good thing to add :)

ArthurSens:
Not sure which one is closer to getting merged, but this PR will have quite a few conflicts with https://github.com/prometheus/prometheus/pull/16951.

We might want to coordinate them

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,501 @@
...
+As mentioned in the [Chunks](#chunks) section, `GaugeType` should still be the counter reset hint/header.
+
+Cumulative counter-like metrics ingested via the OTLP endpoint will also have a `__temporality__="cumulative"` label added.
+
+**Pros**
+* Clear distinction between OTEL delta metrics and OTEL gauge metrics when ingested into Prometheus, meaning users know the appropriate functions to use for each type (e.g. `sum_over_time()` is unlikely to be useful for OTEL gauge metrics).
+* Closer match with the OTEL model - in OTEL, counter-like types sum over events over time, with temporality being an property of the type. This is mirrored by having `__type__="counter"` and `__temporality__="delta"/"cumulative"` labels in Prometheus.
+
+**Cons**
+* Introduces additional complexity to the Prometheus data model.
```
bwplotka, line 299: Aren't gauges with __temporality__ label introduce even more complexity?

Especially when explaining why it's gauge vs counter to an Otel user and when in future we add switch to have it as a counter? 🙈  

Switching to counters with __temporality__ might actually be less complexity, no?

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,501 @@
...
+* Confusing overlap between gauge and delta counters. Essentially deltas already exist in Prometheus as gauges, and deltas can be viewed as a subset of gauges under the Prometheus definition. The same `sum_over_time()` would be used for aggregating these pre-existing deltas-as-gauges and delta counters, creating confusion on why there are two different "types".
```
bwplotka, line 300: > Essentially deltas already exist in Prometheus as gauges, and deltas can be viewed as a subset of gauges under the Prometheus definition.

Is it really true? What exactly is that subset? That gauge can go down and counter not? Counter can do down meaning reset though. Delta can go down because it's a diff. Gauge is for a current value, Delta is for diff. I see equal confusion when used on either type.

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,501 @@
...
+  * Pre-existing deltas-as-gauges could be converted to counters with `__temporality__="delta"`, to have one consistent "type" which should be summed over time.
```
bwplotka, line 301: What do you mean? What if we start with counter type with __temporality__ label?

bwplotka (dismissed):
Thanks for amazing work on this one!!!

**Can I challenge the need of delta as gauge idea? Given new changes, I'd argue we should use counter +__temporality__ from day 1** 

Feel free to dismiss this, but wanted to put some findings here! 🤗 

## Context 

The Initial version of this proposal assumed we DON'T want `__temporality__` label, but instead we will have gauge for now and specific delta type later.

However @aknuds1  pinged me on the PR https://github.com/prometheus/prometheus/pull/16971 that is adding it, so I dived into the newest state of this proposal.

So the current plan is to:
* Add __temporality__ label to gauges  that represents delta
* NOT add a special type in future, but instead ingest them as counters one day.

I would argue that IF we decided to switch direction and **share** existing types and add another __temporality__ dimension, it might be more beneficial to put that dimension on counters and ingest deltas as counters (with __temporality__=delta) from day 1.

## Rationales

1. We wanted to experiment and gather data. So why not starting with counters? We plan to already "confuse/break" people by asking them to ingest delta-gauges and name them as counters,  use counter based functions (increase etc).. but we are afraid to add some unused (see why unused below) metadata to __type__ to counter here?
1. I would challenge the fact we break people given we break the "counter definition" by putting delta as counters (do you have at least one practical example)? Our metadata-wal-records is unusable for long term storages (overhead, not reliable), and none of internal Prometheus code do anything different between gauge vs counter. type and unit will improve this, but it's too new feature to be already adopted. So why not just do this, given experimentation phase? There might be some tools that maybe do something different for `_total` names, but[ from your examples](https://github.com/prometheus/prometheus/pull/16360) I see you planned to add delta as gauges with counter-like naming (!), so we already break this? 🙈 
1. Using counter with new dimension  is more aligned with OpenTelemetry, no?
1. Using counter with new dimension feels aligned with  histogram vs gaugehistogram story no? gauge histogram is NOT delta sum histogram
**1. Switching Otel Prometheus users from delta as gauge to delta as counters will be MUCH more painful later once type-and-unit is adopted AND we get users to use gauge now.**
1. Maybe I get this wrong, but we planned to add delta as gauge. Then we noticed some PromQL functions would work better if we know if it's a normal gauge vs delta gauge, so we planned to add __temporality__ labels to gauges... Wouldn't it be 100x easier if we start PromQL handling with counter + __temporality__ from day 1 here vs trying to remove code later on to handle special deltas with __temporality__?

## To sum up

I believe we have two options:

1. We follow Otel model with counter/histogram + __temporality__. For this I'd argue using gauges now is more damaging than using counters on day 1.
2. We create new model without new label dimension -  with gauge, counter, delta (and gaugehist, hist, deltahist). If we follow this using gauges might make sense tmp.

Rightly so, seems that this proposal proposes (1) - in this case I'd use counters from day 1.

proposals/0048-otel_delta_temporality_support.md
fionaliao: The idea is to treat the otel properties e.g. otel type, monotonicity,  basically as a separate model 
(at some point, I suggested `__otel_temporality__` as the label name rather than just `__temporality__`) that can be used if people wanted. And marking the prometheus type as `gauge` as in general gauge handling would work (e.g. warning not to use rate() or increase() with non-counter type: https://github.com/prometheus/prometheus/pull/16632).

I agree that this can be more complex to explain to users and that's a downside, but it's less complexity for prometheus internally.

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,501 @@
...
+* Systems or scripts that handle Prometheus metrics may be unaware of the new `__temporality__` label and could incorrectly treat all counter-like metrics as cumulative, resulting in hard-to-notice calculation errors.
```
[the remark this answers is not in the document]
fionaliao: > I believe no system that handles Prometheus types because we never expose them.

There is the [metadata api](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata) that exposes type information.

Grafana uses that to add a hint suggesting to use `rate()` for a counter metric (falling back on suffix matching if no metadata):
https://github.com/grafana/grafana/blob/04f392d37be0ec64e76cc924f7b73a95be1a1d18/packages/grafana-prometheus/src/language_provider.ts#L221-L234
https://github.com/grafana/grafana/blob/1cb90cd3b20b2b7ad4161e6e3d5c33e34f7da64e/packages/grafana-prometheus/src/query_hints.ts#L100-L127

For this specific case, if we did function overloading (increasing internal Prometheus complexity), `rate()` could be used for deltas as well, then this wouldn't be a problem.

I wouldn't be surprised if other systems also depended on this typing, but admittedly I don't have other examples

proposals/0048-otel_delta_temporality_support.md
fionaliao: I like `__monotonicity__="monotonic/non-monotonic"` (having descriptive label values is nice)

fionaliao:
> But.. what if we have ct per sample feature literally next week or month? Would it be enough to convince us to use counters (without temporality label?). The mini cumulatives idea make sense to me. If yes then getting CT per sample (which is needed for other reasons too) might be easier together if we "swarm" this vs getting through delta as gauge period, no? (:

Yeah I think if CT-per-sample is available soon then deltas-as-gauge wouldn't be necessary. I would like to retain the temporality label even if we had CT-per-sample, but it would be mostly informational (or useful if we needed to translate back to otel).

However, we have most of the code for deltas-as-gauges ready though, so would it be okay to have that merged while working on CT-per-sample? The delta feature experimental and there are warnings around label changes.

> I don't yet understand the non-monotonic sum from statsd question. Do you mean statsd counter is literally like otel up and down counter or something? I would need to learn more about it here, but it sounds like a monotocity dimension feature separate to delta, no?

Yes, from the statsd spec: "A counter is a gauge calculated at the server. Metrics sent by the client **increment or decrement** the value of the gauge rather than giving its current value." 
Also see this discussion: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/1789
(also counters are subsets of gauges in statsd lol)

It's a separate to temporality, but it's another property of a otel counter

proposals/0048-otel_delta_temporality_support.md
bwplotka: Good question, I think `__monotonic__` or [`__is_monotonic__`](https://github.com/open-telemetry/opentelemetry-proto/blob/8672494217bfc858e2a82a4e8c623d4a5530473a/opentelemetry/proto/metrics/v1/metrics.proto#L227) would be ok choices.

proposals/0048-otel_delta_temporality_support.md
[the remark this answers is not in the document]
bwplotka: Cleaning up, perhaps we should follow main discussion or start new one.

Also see https://github.com/prometheus/proposals/pull/48#event-19403735518

proposals/0048-otel_delta_temporality_support.md
bwplotka: Cleaning up, perhaps we should follow main discussion or start new one.

Also see https://github.com/prometheus/proposals/pull/48#event-19403735518

proposals/0048-otel_delta_temporality_support.md
bwplotka: Cleaning up, perhaps we should follow main discussion or start new one.

Also see https://github.com/prometheus/proposals/pull/48#event-19403735518

bwplotka: Cleaning up, perhaps we should follow main discussion or start new one.

Also see https://github.com/prometheus/proposals/pull/48#event-19403735518

proposals/0048-otel_delta_temporality_support.md
```diff
@@ -0,0 +1,501 @@
...
+
+We decided not to go for this approach for the initial version as it is more invasive to the Prometheus model - it changes the definition of a Prometheus counter, especially if we allow non-monotonic deltas to be ingested as counters. This would mean we won't always be able convert Prometheus delta counters to Prometheus cumulative counters, as Prometheus cumulative counters have to monotonic (as drops are detected as resets). There's no actual use case yet for being able to convert delta counters to cumulative ones but it makes the Prometheus data model more complex (counters sometimes need to be monotonic, but not always). An alternative is allow cumulative counters to be non-monotonic (by setting a `__monontonicity__="false"` label) and adding warnings if a non-monotonic cumulative counter is used with `rate()`, but that makes `rate()` usages more confusing.
+
+We may switch to this option if marking as the type as gauge appears too confusing to users.
```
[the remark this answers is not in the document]
bwplotka: Cleaning up, perhaps we should follow main discussion or start new one.

Also see https://github.com/prometheus/proposals/pull/48#event-19403735518

bwplotka:
Do you mind a quick rebase? We just fixed CI.

ywwg:
The approach in this Proposal have been superseded by https://github.com/prometheus/proposals/pull/77 so we are going to close it to reduce confusion.

PR #52, opened 2025-05-02
Add Extended range selectors proposal

roidelapluie:


proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,396 @@
...
+
+## Q&A
+
+*Why `smoothed` does not apply extrapolation if there are no points before or after the range?*
+
+Extrapolation for newly appearing counters with new labels or disappearing would
+compromise aggregability. Instead, by duplicating the first data point at the
+beginning of the range, we maintain accuracy without introducing artificial data calculations.
+
+For queries at the current time ("now"), we also deliberately avoid extrapolation at the end of the range. While this **will always** show a slight drop for metrics with constant rates and "smoothed", this accurately reflects that the data is incomplete at the range boundary. In future iterations, we could introduce a special annotation to indicate this boundary condition if users find it problematic in practice.
```
free, line 322: `smoothed` works fine in the middle of a time series (i.e. for historical queries). But the fact that it will produce different results in a recording rule vs later on (beyond the usual race conditions that may lead to a couple of missing samples every now and then at the time of rule evaluation) means that it is really only useful for charts and such.

I suppose that's fine, since it's an optional modifier, but I would consider strongly discouraging its use in rules. E.g. if you scrape `instance="A"` with an offset of 1 second and `instance="B"` with an offset of 14 seconds (on a scrape interval of 15 seconds), then `increase(foo{instance="A"})` will be consistently ~30% higher than `increase(foo{instance="B"})` for a `foo` that increases at the same rate on both instances.

Also, `metric smoothed` won't even produce an output in real time, since there's never a next sample at the time of evaluation.

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,396 @@
...
+![](../assets/2025-04-04_extended-range-selectors-semantics/2da740cff0429094d205883be2161df1.png)
+
+*Anchored vectors ensure boundary completeness by including real samples at range boundaries. The diagram shows how a sample just before the range start (outside the original range) is included, and the last sample at range end is duplicated at the end of the range, making sure there are points to cover the complete interval [start, end].*
+
+**Behavior**:
+- This mode treats the selector range as **closed on both ends**: `[start, end]`.
+- At both the start and end of the range (or evaluation time, for instant
+  selectors), Prometheus will search **backward** using a lookback window to find
+  the sample just before the boundary.
+- If there is no **previous sample**, we duplicate the first sample in the range at the start of the range.
```
free, line 214: Actually duplicating the previous sample at range boundary has the side-effect of losing timestamp information. This makes no difference for `increase()` (since it's simply the difference between the samples' values, regardless of timestamp).

But `rate()` will be negatively affected by essentially changing the timestamps of the samples at the two ends. Given a few missing samples (or a range that is not a multiple of the scrape interval) the `rate()` of a time series that increases at a constant rate will jump around (because one of both ends get "stuck", while `rate()` continues to divide the jittery increase by a fixed time interval). This would not be the case if you retained the two samples' original timestamps.

IOW, `increase()` without interpolation will inevitably jitter whenever you have one or more missing samples at your range end. `rate()` doesn't have to.

Side note (and personal point of view): For me, the clearest way to see counters is (as in your data sets above) as a sequence of timestamped value deltas (because that's precisely what the code producing them does: "I've written 5 more bytes"). Unfortunately, due to the limitations of scraping, counters are represented as sequences of monotonically increasing values. But if you ignore the representation and focus strictly on the meaning, then it's a lot easier to reason about what `increase(foo[1m])` or `rate(foo[1m])` should be doing: sum all the value deltas in the given range (and divide them by the sum of their time deltas, in the case of `rate()`).

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,396 @@
...
+- **Abstract away all underlying heuristics**: This is not about hiding complexity—it’s
+  about making trade-offs more explicit and controllable. Users still need to
+  understand how their choice of modifier affects the results, and documentation
+  and naming should reflect that.
+
+## How
+
+### How This Proposal Works
+
+This proposal introduces two new **modifiers for range and instant selectors**
```
charleskorn, line 186: Could you elaborate on why the new modifiers are necessary for instant vector selectors? It seems from the motivation above and discussion below the new modifiers are only necessary for range vector selectors.

proposals/2025-04-04_extended-range-selectors-semantics.md
free: Leaving aside the fact that it shifts the resulting timestamps, AFAICT the `offset` approach is only a best-effort hack. In order for it to work reliably it would have to be something like `offset <lookback_window>`. Something like `offset <scrape_interval>` (or a multiple thereof; also a bit of a hack, BTW, because you'd be hardcoding the scrape interval in your query) breaks if you miss a scrape (or more). Whereas, even with a missing sample(s), the metric did actually experience an increase. So not producing an output under those conditions is very much incorrect behavior.

As for interpolation / extrapolation, I agree with you that they only make up misleading data. E.g. why should it be that a query like `increase(foo[1ms])` would produce a non-zero output when `foo` has ` orders of magnitude lower resolution? How can anyone tell that the underlying metric actually increased within that 1 ms?

IMHO, one should embrace the fact that the data has the resolution that it does, not whitewash it for the sake of smooth lines on a chart (while at the same time, not generating weird artifacts at the edges or around step increases, as the current implementation does).

proposals/2025-04-04_extended-range-selectors-semantics.md
free: As per the above, data has whatever resolution it has. So if in the last 5 minutes I have 3 increases covering a time span of 6 minutes (whether equally spaced or with gaps) why would I add them up and divide by 5 minutes?

IIRC the current implementation does this already: it tries to extrapolate at the ends, but if the samples are too far from the end, then it will adjust the duration that it divides by accordingly. While nominally claiming to be increase divided by time. Because it's true. It's just that said time is not always the 5 minutes specified by the user, because you don't always have 5 minutes worth of data. I don't see anything contentious about that.

proposals/2025-04-04_extended-range-selectors-semantics.md
charleskorn: Makes sense. Perhaps we can start with adding these keywords just for range vector selectors for now?

proposals/2025-04-04_extended-range-selectors-semantics.md
free: [I am easily triggered and I don't like where this discussion is headed. So I'll leave this here and try my best to move along.]

How exactly is summing made up data (which is what interpolation gives you) better than summing data that is shifted by some time delta lower than the scrape interval? What if my counter is an actual step function?

And how is summing "smoothed" data shifted by a full scrape interval (when that happens to work) better than data shifted by less than a full scrape interval? (Because that's the data you have.) And if that is indeed a problem, what's stopping you from giving the resulting value the timestamp of the last sample, so you get zero time shift and zero hallucinated data?

All values output by Prometheus are shifted by some time delta less than a full scrape interval (if everything works perfectly). Why is this suddenly different? Did anyone ever consider smoothing gauges, as an improvement over the status quo? If not, why not?

free: > You would like some kind of "aligned_smoothed" which aligns on the last sample and smooth the beginning of the selector. I will consider that and might update the proposal.

No, I do not want any smoothing at all. IMHO `increase()` should return exactly what it does in your proposed `anchored` mode; `rate()` would be implicitly "smoothed" by dividing that value by the actual duration covered by the samples instead of a fixed `5m` (or whatever the query specifies). No need for interpolation at all.

If you think of a counter as a sequence of value deltas instead of a sequence of absolute values, then it makes perfect sense to think of the increase over the past 5 minutes as being the sum of all value deltas recorded in the past 5 minutes. Even if the first value delta covered e.g. the previous `3m @now-4m37s` because you happened to miss a couple of scrapes just before. This is exactly what your proposed `anchored` does (for `increase`) except that, for whatever reason, it forgets the timestamps along the way.

Now, if you don't mangle the timestamps, then when computing `rate()`, you can simply take the increase (the sum of all the value deltas that fall into the last `5m`) and divide it by the sum of the corresponding time deltas. No need to make up new timestamps and then interpolate / extrapolate to make up for it. Just use the actual data and accept that it will be arbitrarily shifted by a few seconds because of the way scraping and evaluation are spread out over time rather than perfectly synchronized.

free: Smoothing at either end only gives you absurd results, such as "2.375 process restarts over the past 5 minutes".

And desperately hanging on to a fixed time range to divide by, will lead to arbitrarily stretching or compressing the (`anchored`) rate whenever samples are missing or not perfectly equally spaced (consider the example of a counter with an actual rate of increase of 1, but with somewhat randomly spaced samples).

All you can do at that point is to religiously stick with `increase(... anchored)` and `rate(... smoothed)` to try and salvage something half-way sane.

fionaliao: If we used the original sample timestamps as the range to divide by, what should happen if there aren't any values within the lookback window? Would it be better to use the original timestamps of the first and last samples in the range, or the duration of the range vector selector?

Additionally, I think having rate() always being equal to increase() divided by the duration of the range vector selector, no matter what modifiers are used, is be a better user experience, as it's consistent so less surprising.

beorn7 (changes requested):
Thank you very much for doing this.

I generally like it. I'm not so sure about some edge cases, but we can go forward with an implementation and see how it works in practice. (It will be an experimental feature for a while, I assume, so we can iterate on it.)

I see in your Q&A that you decided to ignore staleness markers. Maybe you are right. I'm not so sure, but see above. I would still recommend to mention the fact that we plan to ignore staleness markers more prominently (where I added comments asking what to do with staleness markers).

The big remaining question is if `anchored` should work if there is no sample before the range. I was originally thinking it should only return anything if there is actually a sample before the range (or precisely at the beginning of the range). After reading everything, I'm not so sure anymore. I can imagine that the use case I'm often mentioning ("Give my disk usage growth over the last month" – and then I get the growth over the last hour without noticing) will mostly affect gauges, so `foo - foo offset 30d` will work just fine.

Essentially, I'm willing to give it a go pretty much precisely as you propose it here.

beorn7:
Thanks for addressing my comments. I'm still caught between several events at work and in my usual life, but I will finally be able to tackle my review backlog next week, which will allow me to address the open questions.

In the meantime, it would be good to get some feedback from other interested Prometheus developers. I would much prefer more approvals on this than just mine.

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,390 @@
...
+## Non-Goals
+
+While the proposal aims to improve expressiveness and correctness in range-based
+functions, it deliberately avoids certain directions that have historically led
+to complexity or confusion:
+
+- **Define a one-size-fits-all solution**: The goal is not to find a universally
+  "better" rate function that replaces all others. Different use cases have
+  inherently different trade-offs, and the proposal embraces offering multiple
+  well-defined modes instead of forcing consensus around a single behavior.
```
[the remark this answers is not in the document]
beorn7: I still think it is important to make clear here that we also don't want to add dozens of solutions. Otherwise, people will rightfully demand more keywords, implementing their respective detailed approach.

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,390 @@
...
+```promql
+increase(http_requests_total[5m] anchored)
+```
+
+![](../assets/2025-04-04_extended-range-selectors-semantics/2da740cff0429094d205883be2161df1.png)
+
+*Anchored vectors ensure boundary completeness by including real samples at range boundaries. The diagram shows how a sample just before the range start (outside the original range) is included, and the last sample at range end is duplicated at the end of the range, making sure there are points to cover the complete interval [start, end].*
+
+**Behavior**:
+- This mode treats the selector range as **closed on both ends**: `[start, end]`.
```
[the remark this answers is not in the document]
beorn7: I would like to see improved wording here. By now I would go farther and say the current wording is not just confusing, but outright false. A closed interval means that is at the boundary or within the interval, while this mode explicitly includes samples _before_ the interval.

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,390 @@
...
+- At both the start and end of the range (or evaluation time, for instant
+  selectors), Prometheus will search **backward** using a lookback window to find
+  the sample just before the boundary.
```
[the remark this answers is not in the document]
beorn7: That sounds good for now. Could you add a note that staleness markers are deliberately ignored here?

proposals/2025-04-04_extended-range-selectors-semantics.md
beorn7: Yes, that's what I meant.

Let's go with your approach for now, see also other comment thread about this. (Where I came to the conclusion that the use case where you do not want to include series that only start during the interval is likely only happening with gauges, in which case the offset approach already works.)

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,390 @@
...
+- If there is no **previous sample**, we duplicate the first sample in the range at the start of the range.
+- If no sample is found within the selector range, the result dropped.
```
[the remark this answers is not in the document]
beorn7: This needs fixing. How about this:

```suggestion
- If no sample is found within the left-open and right-closed selector range, no result is returned for the respective series.
```

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,390 @@
...
+```
+
+![](../assets/2025-04-04_extended-range-selectors-semantics/31f9c424e4d61716a64aa3d5a1ce7c44.png)
+
+Smoothed vectors behave as if continuous lines were drawn between samples, utilizing linear interpolation to create a smooth transition between data points. This captures the increase of 11.
+
+**Behavior**:
+- This mode estimates the values at the range’s start and end (or instant
+  evaluation time) by **interpolating** between the nearest known samples **before
+  and after** each boundary.
```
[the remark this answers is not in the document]
beorn7: If staleness markers are ignored (what I assume from what you said above), please explicitly mention it here.

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,390 @@
...
+- If there is no datapoint before the range, the first datapoint inside of the range is duplicated at the beginning of the range.
+- If there is no datapoint after the range, the last datapoint inside of the range is duplicated at the end of the range.
+- The interpolation window is controlled by the **lookback delta**,
```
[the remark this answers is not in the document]
beorn7: Yes, definitely. Otherwise we break the composability, which is one of the most intriguing features of this approach.

To clarify: This is about the case where there is a sample _before_ the range, and there is a sample _after_ the range, but no sample _within_ the range. In this case, you essentially draw a line between those two samples, and whatever small range you have in between those two, you return the increase of that line.

beorn7 (changes requested):
A few comments are still unaddressed, see my follow-ups.

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -0,0 +1,400 @@
...
+increase(http_requests_total[5m] anchored)
+```
+
+![](../assets/2025-04-04_extended-range-selectors-semantics/2da740cff0429094d205883be2161df1.png)
+
+*Anchored vectors ensure boundary completeness by including real samples at range boundaries. The diagram shows how a sample just before the range start (outside the original range) is included, and the last sample at range end is duplicated at the end of the range, making sure there are points to cover the complete interval [start, end].*
+
+**Behavior**:
+
+- This mode includes samples within the range `(start, end]` and additionally looks backward to find one sample before the start time.
```
beorn7, line 215: Important nit:

```suggestion
- This mode includes samples within the range `(start, end]` and additionally looks backward to find one sample at or before the start time.
```

proposals/2025-04-04_extended-range-selectors-semantics.md
```diff
@@ -270,6 +270,7 @@ Smoothed vectors behave as if continuous lines were drawn between samples, utili
 - If there is no datapoint after the range, the last datapoint inside of the range is duplicated at the end of the range.
 - If there are no datapoints inside the range but samples exist before and after, it interpolates between those samples to estimate values at the range boundaries.
 - The interpolation window is controlled by the **lookback delta**,
+- If the range selector duration is shorter than the evaluation step interval, the calculation will use the step interval as the effective range duration. A warning will then be returned to the user.
```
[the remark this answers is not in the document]
beorn7: Sure. “Explicit is better than implicit.” And IIRC, we wanted to introduce `step()` anyway. (But it would default to zero in case of instant queries, I guess. Which might be fine in this case.)

So let's do that and not do any implicit "use the step as effective range duration if the range is 'too short'" and not do an empty `[]`?

(I should also mention that we can always add those things later during the experiment. I would prefer to start with a smaller set of changes and take it from there.)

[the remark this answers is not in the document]
beorn7: Fine with me. (But we agree that we should not auto-increase the range selector duration to the step, don't we?)

beorn7 (approved):
I'm fine with the current state. As a lot of work is already ongoing and none of the team members have raised concerns so far, I would recommend merging this rather sooner than later to get out of this "formally still not approved" state. (The design doc does not have to represent the final implementation. In fact, most non-trivial design docs don't. We can still iterate on implementation details.)

bwplotka (approved):
Thanks for this; I'm supportive, LGTM

Thanks for considering [the CreatedTimestamp feature](https://github.com/prometheus/proposals/blob/7f651920c12305ab299c0b50452e719d860f1e97/proposals/2025-04-04_extended-range-selectors-semantics.md#:~:text=What%20is%20the%20impact%20of%20Created%20Timestamp%3F) in the design!

@realschwa would like to double check additionally from the GCM standpoint?

juliusv (commented):
Just took a first look at this as well. Overall the approach seems well thought through and reasonable, good work! A few editorial comments...

ringerc:
Regarding

> What about influence from datapoints far outside the query range?

Are ranges always both left-anchored and right-anchored with the `anchored` keyword? My reading suggests so.

How far will an `anchored` range look back and/or forward for the anchoring point(s)? The delta lookback (staleness threshold)?

What happens if no anchoring point exists in the series? Will the behaviour modifications discussed for anchored series still apply when no anchoring point is found?

frittentheke:
Am I right in reading that the observations about handling of e.g. a single missing data point (https://github.com/prometheus/prometheus/discussions/16891) is also covered in this proposal here?

PR #56, opened 2025-07-15
proposal: TSDB Index Lookup Planning

dimitarvdimitrov:
This proposal introduces extension points for TSDB index lookups that allow different execution strategies to address the problem of inefficient index lookup usage.

Part of the action items are being addressed in https://github.com/prometheus/prometheus/pull/16835


dimitarvdimitrov:
i won't have time to continue working on this through reviews

get in touch with @chencs if you still want this to get in. Some preliminary results from this are in https://github.com/grafana/mimir/issues/11916#issuecomment-3632755773

PR #60, opened 2025-09-09
proposal: TSDB Support for Start Timestamp (ST)

bwplotka:
As discussed in various places (e.g. https://github.com/prometheus/prometheus/pull/17036#issuecomment-3241082314 and delta WG) we decided to create a formal proposal on how CT/ST native Prometheus storage could look like and how to make it useful (unblock) delta temporality.


ywwg (commented):
Thank you for this proposal! I added a bunch of comments, some of which are answered by the paragraph right after the comment 😅 . I think my main concern is nailing down the Goals section.  This is not at all to question whether we should do the work, just that I think our statement of intent needs to be unequivocable.

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+semantics, we could rename to "Start Time/Timestamp" (ST) naming in all elements.
+
+Pros:
+* Consistency with OpenTelemetry and Google
+* There are more future users than previous users, explaining differences between CT and ST.
+
+Cons:
+* Technically, one could argue that both CT and ST naming are correct in this context. Prometheus is cumulative-first system, which means
+  using CT would make sense.
+* We already have CT in scrape protocols and ~Remote Write 2.0; do we need to switch them too?
```
bwplotka, line 484:
```suggestion
* We already have CT in scrape protocols (we go deeper into this in [OM 2.0](https://github.com/prometheus/docs/pull/2636)) and ~Remote Write 2.0; do we need to switch them too?
```

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+* Changing names often creates confusion
+* Having a slight changed name already makes it clear that we talk about Prometheus semantics of the same thing
+
+**Rejected** due to not enough arguments for renaming (feel free to challenge this!).
```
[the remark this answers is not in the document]
bwplotka: Yes! Thanks for mentioning. From notes and memory I think it was your vote to rename vs others to not rename or no opinion. 

This is a starting point for discussion, we can argument otherwise. Do you have more arguments towards renaming that would help motivate us?

When writing this I think I got sold on not renaming. Both are technically correct and given Prometheus being cumulative native it only makes sense to leave like this. This does not mean we shouldn't go against my vote if we have consensus here to rename.  How to find this consensus? (: What would convince you to keep `CT` naming? (: 

Do you mind helping with the vote if you are passionate to challenge this? What are we missing here?

bwplotka:
Back from some PTO/leave, will try to address comments and finalize interface and TSDB piece soon

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+    * [Appender](https://github.com/prometheus/prometheus/pull/17104)
+    * [chunkenc.Iterator](https://github.com/prometheus/prometheus/pull/17176)
+
+* **Other docs or links:**
+  * [PROM-29 (Created Timestamp)](https://github.com/prometheus/proposals/blob/main/proposals/0029-created-timestamp.md)
+  * [Delta type proposal](https://github.com/prometheus/proposals/pull/48), [Delta WG](https://docs.google.com/document/d/1G0d_cLHkgrnWhXYG9oXEmjy2qp6GLSX2kxYiurLYUSQ/edit)
+
+> TL;DR: We propose to extend Prometheus TSDB storage sample definition to include an extra int64 that will represent the cumulative created timestamp (CT) and, for the future delta temporality ([PROM-48](https://github.com/prometheus/proposals/pull/48)), a delta start timestamp (ST).
+> We propose introducing persisting CT logic behind a single flag `ct-storage`.
+> Once implemented, we propose to deprecate the `created-timestamps-zero-injection` experimental feature.
```
ArthurSens, line 23:
```suggestion
> Once implemented, we propose to remove the `created-timestamps-zero-injection` experimental feature.
```

Do we need to go through a deprecation notice for feature flags? I have the impression that the point of feature flags is that they are safe to remove 🤔

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+
+* Remote storages (Remote Write 2.0) (e.g. OpenTelemetry, Chronosphere, Google).
+* PromQL and other read APIs (including federation) (e.g. increased cumulative based operation accuracy).
+
+Furthermore, it would unblock future Prometheus features for wider range of monitoring cases like:
+
+* Delta temporality support.
+* UpAndDown counter (i.e. not monotonic counters) e.g. StatsD.
+
+On top of that this work also allow to improve some existing features, notably various edge cases for native histogram resets hints (e.g. [TSDB chunk merges](https://github.com/prometheus/prometheus/issues/15346)). Arguably, CT can replace resent hint field in native histogram eventually.
```
ArthurSens, line 46:
```suggestion
On top of that this work also allow to improve some existing features, notably various edge cases for native histogram resets hints (e.g. [TSDB chunk merges](https://github.com/prometheus/prometheus/issues/15346)). Arguably, CT can replace reset hint field in native histogram eventually.
```

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+    * For a cumulative, CT SHOULD:
+      * Be the same across all samples for the same count:
+
+      ```
+      CT[0], T[0]
+      CT[0], T[1]
+      CT[0], T[2]
+      ```
+
+  * For delta, CT SHOULD (`=` because it looks the start time is exclusive):
```
ArthurSens, line 123:
```suggestion
  * For delta, CT SHOULD (`=` because it looks like the start time is exclusive):
```

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+
+Native CT support may feel like a big change with a lot of implications. However, we have all the technical pieces to do this, especially if we implement CT with
+(more or less) the same guarantees and details as normal sample timestamp (T), just optional.
+
+General decisions and principles affecting smaller technical parts:
+
+1. We propose to stick to the assumption that CT can change for every sample and don't try to prematurely optimize for the special cumulative best case. Rationales:
+
+* This unlocks the most amount of benefits (e.g. also delta) for the same amount of work, it makes code simpler.
+* We don't know if we need special cumulative best case optimization (yet); also it would be also for some "best" cases. Once we know we can always add those optimizations.
```
ArthurSens, line 183:
```suggestion
* We don't know if we need special cumulative best case optimization (yet); also, it would be for some "best" cases. Once we know we can always add those optimizations.
```

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+
+Let's go through all the areas that would need to change:
+
+### Feature flag `ct-storage`
+
+To develop and give experimental state to users, we propose to add a new feature flag `created-timestamp-storage`. Similar to [exemplar-storage]https://prometheus.io/docs/prometheus/latest/feature_flags/#exemplars-storage) it will tell Prometheus to ingest CT and use new (potentially breaking compatibility) storage formats.
+
+We propose to have a single flag for both WAL, Block storage, etc., to avoid tricky configuration.
+
+Notably, given persistence of this feature, similar to example storage, if users enabled and then disabled this feature, users will might be able to access their CTs through all already persistent pieces e.g. WAL).
```
ArthurSens, line 201:
```suggestion
Notably, given persistence of this feature, similar to examplar storage, if users enable and then disable this feature, users might be able to access their CTs through all already persistent pieces e.g. WAL).
```

ArthurSens (commented):
I'm adding just a few stylish/correction comments. I'm midway through the document and still haven't reviewed the proposed interfaces.

Regarding the CT vs ST discussion, I see the point that we'll always have more new users than old ones, but I feel like the CT terminology is so ingrained in the ecosystem that even if we change it now, people will continue to call it CT. Of course this is based on "voices in my head" and there's no real confirmation that this is gonna happen in the future 😅

proposals/0060-ct-storage.md
ywwg: experimental features esp have no guarantee of stability, so that seems correct

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+* Expanded Scrape parser interface to return `CreatedTimestamp` per sample (aka per line).
+* Optimized Protobuf and OpenMetrics parsers for CT use (TODO links).
+* Implemented an opt-in, experimental [`created-timestamps-zero-injection`](https://prometheus.io/docs/prometheus/latest/feature_flags/#created-timestamps-zero-injection) feature flag that injects fake sample (V: 0, T: CT).
+* Included CT in [Remote Write 2 specification](https://prometheus.io/docs/specs/prw/remote_write_spec_2_0/#ioprometheuswritev2request).
+
+### Background: Delta temporality
+
+See the details, motivations and discussions about the delta temporality in [PROM-48](https://github.com/prometheus/proposals/pull/48).
+
+The core TL;DR relevant for this proposal is that the delta temporality counter sample can be conceptually seen as a "mini-cumulative counter". Essentially delta is a single-sample (value) cumulative counter for a period between (sometimes inclusive sometimes exclusive depending on a system) start(ST)/create(CT) timestamp and a (end)timestamp (inclusive).
```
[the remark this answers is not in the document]
bwplotka: It's mentioned "a delta start timestamp (ST)." in TL;DR, but you are right we could mention it clearly somewhere. Especially once we decide if we go with ST or CT fully, we could just use one and only have a quick mention about other. 👍🏽

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+  * Given immutability invariant (e.g. Prometheus), you can't effectively inject CT at a later time (out of order writes are sometimes possible, but expensive, especially for a single sample to be written in the past per series).
+  * It's prone to OOO false positives (we ignore this error for CTs now in Prometheus).
+  * It's producing an artificial sample, which looks like it was scraped.
+* We can't implement delta temporarily effectively.
+
+## Goals
+
+* [MUST] Prometheus can reliably store, query, ingest and export cumulative created timestamp (CT) information (long term plan for [PROM-29](https://github.com/prometheus/proposals/blob/main/proposals/0029-created-timestamp.md#:~:text=For%20those%20reasons%2C%20created%20timestamps%20will%20also%20be%20stored%20as%20metadata%20per%20series%2C%20following%20the%20similar%20logic%20used%20for%20the%20zero%2Dinjection.))
+* [SHOULD] Prometheus can reliably store, query, ingest and export delta start time information. This unblocks [PROM-48 delta proposal](https://github.com/prometheus/proposals/pull/48). Notably adding delta feature later on should ideally not require another complex storage design or implementation.
```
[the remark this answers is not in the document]
bwplotka: Where exactly in the current wording we have circular must goal?

The reason why I initially put SHOULD is that I can imagine a potential (low probability) reality where the solution for CT storage for cumulatives is considerably different than delta.

When that happens what we do? If we put MUST for deltas, we should NOT go this path, even if it makes sense to split solutions? That's why, from the perspective of finding the best solution for Prometheus, it feels it should be SHOULD, so balance and join solutions if we can.

However, et's change to MUST to prioritize delta -- sounds like the main reason to have MUST is to justify this effort for Grafana goals, and that's ok. 👍🏽 

Want to make that commit?

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+## Non-goals
+
+In this proposal we don't want to:
+
+* Expand details on delta temporality. For this proposal it's enough to assume about delta, what's described in the [Background: Delta Temporality](#background-delta-temporality).
+* Expand or motivate the non-monotonic counter feature.
+
+## How
+
+Native CT support may feel like a big change with a lot of implications. However, we have all the technical pieces to do this, especially if we implement CT with
```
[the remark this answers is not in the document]
bwplotka: Yes, discussed at the bottom and sooner we decide the better (: Resolving this comment.

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+
+3. We propose to treat CT as an *optional* data everywhere, simply because it's new and because certain metric type do not need it (gauges). For efficiency and consistency with scrape and Remote Write protocols we treat default value `int64(0)` as a "not provided" sentinel. This has a consequence of inability to provide the CT of an exact 0 value (unlikely needed and if needed clients needs to use 1m after or before.
+
+4. Given the [CT and ST context](#background-ct-cumulative-vs-st-delta) the CT vs ST meaning blurs. We could rename CT feature now to ST to be consistent with other systems. We propose to stick to "created timestamp" (CT) naming in Prometheus instead of attempting to rename some or all uses to "start timestamp" (ST). The main reason being that CT and ST naming is equally "correct" on storage and given Prometheus is cumulative-first system, we stick to CT. See (and challenge!) the rationales in the [alternative](#rename-ct-to-st).
+
+Given (4) decision, below proposal only uses "CT" naming, but it also means "ST" in delta temporality context.
```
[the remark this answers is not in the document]
bwplotka: Duplicate discussion, I acknowledge you are a bit upset for this first choice.

Let's discuss on https://github.com/prometheus/proposals/pull/60/files#r2344682987

bwplotka (commented):
I addressed first pass, thanks for reviews!

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+
+[Previous section](#background-delta-temporality) argues that conceptually the Cumulative Created Timestamp (CT) and Delta Start Timestamp (ST) are essentially the same thing. This is why typically they are stored in the same "field" in other system APIs and storages (e.g. start time in OpenTelemetry TODO link).
+
+The notable difference when this special timestamp is used for cumulative vs delta samples is the dynamicity **characteristics** of this timestamp.
+
+* For the cumulative we expect CT to change on every new counter restart, so:
+  * Average: in the order of ~weeks/months for stable workloads, ~days/weeks for more dynamic environments (Kubernetes).
+  * Best case: it never changes (infinite count) e.g days_since_X_total.
+  * Worse case: it changes for every sample.
+* For the delta we expect CT to change for every sample.
```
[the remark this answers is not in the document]
ywwg: @fionaliao Haven't we heard of cases where a delta doesn't have a CT? My understanding is that situation exists in the wild whether we want it to or not.  (Is it too late to drop deltas with no CT because that would break people?)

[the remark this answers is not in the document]
vpranckaitis: Even without CTs, deltas could be used and produce useful results with the `sum_over_time()` function. In my opinion, dropping such values might be a step towards "perfect is the enemy of good" territory. 

Maybe a compromise could be to return a warning/info from a rate-like function about missing CTs and potentially inaccurate results?

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,409 @@
...
+  * start_timestamp SHOULD be provided for metrics that follow counter semantics (e.g. counters and histograms).
+  * Receivers MAY reject those series without start_timestamp being set.
+
+* Otel) More detailed, but still descriptive SHOULD rule only: In OpenTelemetry data model ([temporality section](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums:~:text=name%2Dvalue%20pairs.-,A%20time%20window%20(of%20(start%2C%20end%5D)%20time%20for%20which%20the,The%20time%20interval%20is%20inclusive%20of%20the%20end%20time.,-Times%20are%20specified)) ST is generally optional, but strongly recommended. Rules are also soft and we assume they are "SHOULD" because there is section of [handling overlaps](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#overlap). However, it provides more examples, which allow us to capture some specifics:
+  * Time intervals are half open `(ST, T]`.
+  * Generally, ST SHOULD:
+    * `ST[i] < T[i]` (equal means unknown)
```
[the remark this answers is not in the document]
ywwg: does Prometheus have a general philosophy of "fix on ingest" versus "fix at query time"? From my own knowledge, I feel like usually Prometheus does "reject bad data on ingest" as its usual pattern. My concern is that if there's a mix of approaches, it will make Prometheus' behavior harder to reason about. 

As I think about it, I can't think of a problem with force-setting ST=0 if ST=T.

proposals/0060-ct-storage.md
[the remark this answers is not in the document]
bwplotka: Changed to MUST, since it's obviously what happened (:

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+
+Native CT support may feel like a big change with a lot of implications. However, we have all the technical pieces to do this, especially if we implement CT with
+(more or less) the same guarantees and details as normal sample timestamp (T), just optional.
+
+General decisions and principles affecting smaller technical parts:
+
+1. We propose to stick to the assumption that CT can change for every sample and don't try to prematurely optimize for the special cumulative best case. Rationales:
+
+* This unlocks the most amount of benefits (e.g. also delta) for the same amount of work, it makes code simpler.
+* We don't know if we need special cumulative best case optimization (yet); also it would be also for some "best" cases. Once we know we can always add those optimizations.
```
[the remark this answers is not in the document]
bwplotka: Yes, this is somewhat possible in case of errors or somebody duplicates streams of samples. Some backends override the old data that overlaps. Ideally we allow downstream backends to decide what to do, but for Prometheus we decided to not imply any validation by default at the moment and handle things on query time. 

All noted in `### Proposed ST semantics and validation`

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+
+### Proposed CT semantics and validation
+
+See the [official CT semantics in the ecosystem](#background-official-ctst-semantics). In this section we propose what semantics we recommend (or enforce).
+
+CT/ST notion was popularized by OpenTelemetry and early experience exposed a big challenge: CT/ST data is extremely unclean given early adoption, mixed instrumentation support and multiple (all imperfect) ways of ["auto-generation"](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstarttimeprocessor) (`subtract_initial_point` might the most universally "correct", but it added only recently). This means that handling (or reducing ) CT errors is an important detail for consumers of this data.
+
+TODO: Just a draft, to be discussed.
+TODO: There are questions around:
+* Should we do inclusive vs exclusive intervals?
```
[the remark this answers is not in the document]
bwplotka: I'm leaving this open, but we agreed to keep the semantics out of this proposal. See ### Proposed ST semantics and validation

Given majority of our delta data is coming from OTel I'd lean towards Otel semantics, so `exclusive`

https://opentelemetry.io/docs/specs/otel/metrics/data-model/#:~:text=Delta%20temporality%20means%20that%20successive%20data%20points%20advance%20the%20starting%20timestamp.%20For%20example%2C%20from%20start%20time%20T0%2C%20delta%20data%20points%20cover%20time%20ranges%20(T0%2C%20T1%5D%2C%20(T1%2C%20T2%5D%2C%20(T2%2C%20T3%5D%2C%20and%20so%20on.

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,496 @@
...
+* Given optionality of this feature, can we even reject sample on TSDB Append if CT is invalid? (MUST or SHOULD on interface?)
```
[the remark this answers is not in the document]
bwplotka: Yes, all noted as so

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,409 @@
...
+  * start_timestamp SHOULD be provided for metrics that follow counter semantics (e.g. counters and histograms).
+  * Receivers MAY reject those series without start_timestamp being set.
+
+* Otel) More detailed, but still descriptive SHOULD rule only: In OpenTelemetry data model ([temporality section](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums:~:text=name%2Dvalue%20pairs.-,A%20time%20window%20(of%20(start%2C%20end%5D)%20time%20for%20which%20the,The%20time%20interval%20is%20inclusive%20of%20the%20end%20time.,-Times%20are%20specified)) ST is generally optional, but strongly recommended. Rules are also soft and we assume they are "SHOULD" because there is section of [handling overlaps](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#overlap). However, it provides more examples, which allow us to capture some specifics:
+  * Time intervals are half open `(ST, T]`.
+  * Generally, ST SHOULD:
+    * `ST[i] < T[i]` (equal means unknown)
```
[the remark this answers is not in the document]
bwplotka: We agree to deal on query time for now. For ST==T currently proposal proposes to switch to 0 ("unknown") or at least treat it as so

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,404 @@
...
+
+* Remote storages (Remote Write 2.0) (e.g. OpenTelemetry, Chronosphere, Google).
+* PromQL and other read APIs (including federation) (e.g. increased cumulative based operation accuracy).
+
+Furthermore, it would unblock future Prometheus features for wider range of monitoring cases like:
+
+* Delta temporality support.
+* UpAndDown counter (i.e. not monotonic counters) e.g. StatsD.
+
+On top of that this work also allow to improve some existing features, notably various edge cases for native histogram resets hints (e.g. [TSDB chunk merges](https://github.com/prometheus/prometheus/issues/15346)). Arguably, ST can replace resent hint field in native histogram eventually.
```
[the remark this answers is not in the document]
bwplotka: done, thanks!

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,404 @@
...
+
+Notably, given persistence of this feature, similar to example storage, if users enabled and then disabled this feature, users will might be able to access their STs through all already persistent pieces e.g. WAL).
+
+This feature could be considered to be switched to opt-out, only after it's finished (this proposal is fully implemented) stable, provably adopted and when the previous LTS Prometheus version is compatible with this feature.
+
+### Proposed ST semantics and validation
+
+See the [official ST semantics in the ecosystem](#background-official-st-semantics). In this section we propose what semantics we recommend (or enforce).
+
+ST notion was popularized by OpenMetrics and OpenTelemetry. Early experience exposed a big challenge: ST data is extremely unclean given early adoption, mixed instrumentation support and multiple (all imperfect) ways of ["auto-generation"](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstarttimeprocessor) (`subtract_initial_point` might the most universally "correct", but it added only recently). This means that handling (or reducing ) ST errors is an important detail for consumers of this data.
```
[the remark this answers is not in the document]
bwplotka: done, thanks!

proposals/0060-ct-storage.md
```diff
@@ -0,0 +1,404 @@
...
+
+Similar V2 records will be created for Histogram types.
+
+#### TSDB blocks
+
+ST will be persisted in TSDB blocks using a new chunk encoding currently under development.
+
+#### Memory Snapshots and Head Chunks
+
+There will be likely some code to add to ensure STs are used properly, but the new proposed chunk format from [TSDB Block](#tsdb-blocks) section, should immidately work with the [Memory Snapshot](https://github.com/prometheus/prometheus/blob/747c5ee2b19a9e6a51acfafae9fa2c77e224803d/tsdb/docs/format/memory_snapshot.md) and [Head Chnks](https://github.com/prometheus/prometheus/blob/747c5ee2b19a9e6a51acfafae9fa2c77e224803d/tsdb/docs/format/head_chunks.md) formats.
```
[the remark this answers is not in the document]
bwplotka: done, thanks!

bwplotka:
Addressed all comments, plus it's already implemented 🎉 

Waiting for formal approvals as discussed in Slack

vpranckaitis (approved):

ywwg (approved):
🚀

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,409 @@
...
+
+2. Similarly, we propose to not have new special ST storage cases per metric types beyond the existing difference between native histograms and float metrics. Some systems allow optional STs on gauges (e.g. OpenTelemetry). We propose keeping that storage flexibility.
+
+3. We propose to treat ST as an *optional* data everywhere, simply because it's new and because certain metric type do not need it (gauges). For efficiency and consistency with scrape and Remote Write protocols we treat default value `int64(0)` as a "not provided" sentinel. This has a consequence of inability to provide the ST of an exact 0 value (unlikely needed and if needed clients needs to use 1m after or before.
+
+4. Let's go through all the areas that would need to change:
+
+### Feature flag `st-storage`
+
+To develop and give experimental state to users, we propose to add a new feature flag `st-storage`. Similar to [exemplar-storage]https://prometheus.io/docs/prometheus/latest/feature_flags/#exemplars-storage) it will tell Prometheus to ingest ST and use new (potentially breaking compatibility) storage formats.
```
dashpole:
```suggestion
To develop and give experimental state to users, we propose to add a new feature flag `st-storage`. Similar to [exemplar-storage](https://prometheus.io/docs/prometheus/latest/feature_flags/#exemplars-storage) it will tell Prometheus to ingest ST and use new (potentially breaking compatibility) storage formats.
```
Link formatting

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,409 @@
...
+
+We propose to have a single flag for both WAL, Block storage, etc., to avoid tricky configuration.
+
+Notably, given persistence of this feature, similar to example storage, if users enabled and then disabled this feature, users will might be able to access their STs through all already persistent pieces e.g. WAL).
```
dashpole:
```suggestion
Notably, given persistence of this feature, similar to example storage, if users enabled and then disabled this feature, users will might be able to access their STs through all already persistent pieces (e.g. WAL).
```

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,409 @@
...
+
+We propose to not validate ST values on Prometheus write (Appender level). ST is treated as opaque and optional element of the sample. Prometheus will store whatever int64 value it receives, with `0` meaning **not provided**. This is a deliberate design choice, given:
+* ST data in Prometheus ecosystem is a new concept. ST is expected to be missing in many cases (e.g. old SDKs, old exporters, exporters where ST is non-trivial to detect).
+* ST in wider ecosystem is often unclean (mixed instrumentation, imperfect auto-generation).
+* The exact consumption semantics is still experimental thus we want to stay flexible and don't block future use cases (e.g. exact semantics of ST > T).
+* We can always add validation features opt-in.
+
+For unknown start-time reset points (e.g. OpenTelemetry cumulative points where `ST == T`), ingestion layers (such as OTLP receivers or Remote Write endpoints) are expected to map these to `0` (unknown ST) when appending to storage. This optimizes storage (0 takes exactly 1 bit in chunk/WAL formats) and simplifies detection without requiring timestamp matching.
```
dashpole: What about the points that follow in the series? Won't those have wildly incorrect rates?

dashpole (approved):

proposals/0060-st-storage.md
bwplotka: Yea, I thought we decided this and AI filled this -  but it's good to unpack again. Let's stop and decide.

~Benefit of keeping ST==T vs ST==0 separate is that you know if unknown ST is on cumulative vs delta. But is that separation needed? Do we want that?~

~Naive thinking... For cumulatives, it feels it should resolve to old value detection. For delta it should treat ST==T as ST is previous T, right? Unless we detect delta vs cumulatives some other way~

EDIT: I read this as delta ST==T is converted to 0, but it's for cumulatives which we kind of have to support as 0 anyway? Why not making those consistent?

cc @vpranckaitis @krajorama what we assume on current query path? I have to catch with the latest updates there (:

bwplotka: Actually it's only for cumulatives here... so could be ok?

It's just weird that on scrape unknown ST is 0 and from OTLP it's ST==T. We need to handle 0 anyway.

Deltas are allowed to have ST=T so this section might be what we want?

dashpole: So for context, the ST == T points come from this: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#cumulative-streams-inserting-true-reset-points.

Consider:

* Point 1 Has `ST == T1`.  This seems OK to set start time to unknown.
* Point 2 has `ST == T1`, so interval is [T1, T2].  This seems problematic because the interval is a valid time interval.  But OTel sends the entire cumulative value with this point, so rates are way too high.

Ideally, we would somehow mark `T1` as permanently unknown for the series, so that Point 2 would also get an unknown timestamp?

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,409 @@
...
+
+## Why
+
+The main goal of this proposal is to make sure [PROM-29's created timestamp (CT)](0029-created-timestamp.md) information is reliably and efficiently stored in Prometheus TSDB, under the new chosen start timestamp (ST) name. This means ST can be:
+
+* Written via TSDB Appender interfaces.
+* Query-able via TSDB Querier interfaces.
+* Persistent in WAL.
+* Watch-able (WAL) by Remote Writer.
+* (eventually) Persistent in TSDB block storage.
```
krajorama: We put it into blocks and handle it when compacting (head) blocks, so this should work - if not, it's a bug.

```suggestion
* Persistent in TSDB block storage.
```

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Background: ST feature
+
+[PROM-29](0029-created-timestamp.md) introduced the created timestamp (CT) concept for Prometheus cumulative metrics. Since then [the community adopted the Start Timestamp name](https://github.com/prometheus/prometheus/issues/17416).
+
+Semantically, ST represents the time when "counting" (from 0) started. In other words, ST is the time when the counter started counting from zero.
+
+Conceptually, ST extends the Prometheus data model for cumulative monotonic counters as follows:
+
+* (new) int64 Timestamp (ST): When counting started.
```
krajorama: nit:

```suggestion
* (new) int64 Start Timestamp (ST): When counting started.
```

proposals/0060-st-storage.md
krajorama: Assuming we store ST==T as is, there are two cases at query and two ways we want to use the start time:

- the first sample is in range:
  - reset detection (works on pairs of samples): we check that ST is between sample 2 and 1 and we can also check if it was unknow at the 1st sample, so we can avoid detecting the reset
  - extrapolation: no pair to check, ignored
- the first sample is not in range
  - reset detection: we don't do it for the left side of the range
  - extrapolation: I think since the first sample wasn't in range, the zero point calculated from ST wouldn't be in range either??? so not relevant?

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,409 @@
...
+
+Similar V2 records will be created for Histogram types.
+
+#### TSDB blocks
+
+ST will be persisted in TSDB blocks using a new chunk encoding. The idea is to combine forces with the @roidelapluie initiative
+to optimize XOR chunk around joint DoD encoding for both timestamps and value change semantics for common scenarios.
+
+[XOR2 encoding details](https://github.com/prometheus/prometheus/blob/e793b26713cc7052c7558ae6ceffaa66c2a5b39f/tsdb/docs/format/chunks.md#xor2-chunk-data)
+
```
krajorama, line 425: Missing reference to work on native histograms storage. https://github.com/prometheus/prometheus/pull/18609

krajorama (commented):
Looking good, I think the unknown start time is the sticking point.

proposals/0060-st-storage.md
bwplotka: Ah, ok, important context.

Sounds like @dashpole, you propose to optimize for another algorithm for unknown ST handling ([`true_reset_point`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstarttimeprocessor#strategy-true-reset-point) vs the one we have now in Prometheus with `st-synthesis`, so [`subtract_initial_point `](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstarttimeprocessor#strategy-subtract-initial-point) that arguably fits well with the traditional rate logic).

1. `true_reset_point` never worked in Prometheus, does it make sense to optimize for it now vs recommending to not use it?
2. Can we even handle `true_reset_point`? `the first sample is in range` case even.. what do we do if we see 


```
(ST1, T2),(ST3,T3) aka "true reset point", (ST3,T4)
```

Between T2 and T3 we have zero idea if there was a reset or not, we just know the source didn't know either, no? am I missing sth?

3. What if we some backend designed another one? Can we even keep up with different variations?

After all Prometheus is some backend and we could be opinionated in minimal semantics we need 🤔

dashpole: My main point was that I don't think we can't make the start timestamp of _only the first point_ unknown without making all of the start timestamps after it also unknown.

I would suggest: If this case causes problems, reject points with ST == T. If it doesn't cause problems, just accept them as-is. I couldn't quite tell from @krajorama's comment above if it will actually cause issues or not.

I don't think we should introduce special handling for this case.

vpranckaitis: > My main point was that I don't think we can't make the start timestamp of only the first point unknown without making all of the start timestamps after it also unknown.

This isn't a problem for query side, and we still can recognize OTel's cumulative with unknown start time even if we modify just the first ST in the sequence.

image

The image above shows how start timestamps should behave in OTel for deltas and cumulative counters. If we would compare datapoints `T1` and `T2` between delta and cumulative with unknown start time, we would see that the only difference is the value of `ST1`. However, the treatment of these two should be different: there should be a reset between `T1` and `T2` for deltas, but no reset for cumulative with unknown start time. In order to decide, we check that `ST2 == T1` to rule normal cumulative, and then we check that `ST1 == T1` to single out cumulative with unknown start time. 

Datapoints `T3` and beyond don't matter, because for deltas you can apply the same logic as in previous paragraph, and for cumulatives `ST3 < T2`.

Now what if we decided to replace `ST1 == T1` with some other value, let's say 0, but keep all the other ST values intact? Then you would still check `ST2 == T1` to rule out normal cumulative, and then you would check `ST1 == 0` to single out cumulative with unknown start time. You could even come up with some other magic number for unknowns, and then you could check `ST1 == math.MinInt64` or whatever to single out cumulative with unknown start time case.

Or to explain this in a different way: to known whether a datapoint `X` belongs to a cumulative series with unknown start time, you look whether there's a datapoint at time `ST_X`, and check if it that datapoint has unknown start time. This works in reverse: if a datapoint `Y` has unknown start time, then any datapoint that has `ST = T_Y` also belong to the same cumulative series with unknown start time. Thus you can change only the datapoint `Y`, and the change could be traced back from any other datapoint in this cumulative series, as if the change has propagated to the whole cumulative series.

---

Now, should we commit to doing such replacement? In my opinion, if we don't come up with a very good reason why it would be useful to treat `ST==T` case as a reset, then we shouldn't do that. Otherwise, we're introducing complexity for little benefit.

ywwg:
Per our meeting today, we will be updating this proposal with some clarifications and additions:

- [x] We will not be deprecating start time injection
- [x] But we will rename it and/or change how it is configured to work with st-per-sample
- [x] We will add detail to the proposal about how those features will interact (e.g., zero-injected st takes precedence)

ywwg:
ok I merged in the last changes talking about zero injection, are we good to merge?

bwplotka:
I found some nits on https://github.com/prometheus/proposals/pull/87/changes#r3501661834 also curious about https://github.com/prometheus/proposals/pull/60#discussion_r3247128155 - even if we don't plan to decide, let's state so (I might have lost context, need to check what's the latest 🙈 )

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,420 @@
...
+* For the cumulative we expect ST to change on every new counter restart, so:
+  * Average: in the order of ~weeks/months for stable workloads, ~days/weeks for more dynamic environments (Kubernetes).
+  * Best case: it never changes (infinite count) e.g days_since_X_total.
+  * Worse case: it changes for every sample.
+* For the delta we expect ST to change for every sample.
+
+### Background: Official ST semantics
+
+There are certain semantic constraints on ST values to make them useful for consumers of the known metric types and semantics. Different systems generally have similar semantics, but details and how strong (or practically enforced) those rules are may differ. Let's go through state-of-the-art, just to have an idea what we could propose for Prometheus ST enforcement (if anything):
```
krajorama: nit: there's no way to avoid proposing something as we have to write the code and it should be deterministic.

```suggestion
There are certain semantic constraints on ST values to make them useful for consumers of the known metric types and semantics. Different systems generally have similar semantics, but details and how strong (or practically enforced) those rules are may differ. Let's go through state-of-the-art, just to have an idea what we could propose for Prometheus ST enforcement:
```

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,420 @@
...
+    * For a cumulative, ST SHOULD:
+      * Be the same across all samples for the same count:
+
+      ```
+      ST[0], T[0]
+      ST[0], T[1]
+      ST[0], T[2]
+      ```
+
+  * For delta, ST SHOULD (`=` because it looks the start time is exclusive):
```
krajorama: nit: sentence doesn't make sense to me (non native speaker)

```suggestion
  * For delta, ST SHOULD (`=` because time intervals are half open):
```

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,420 @@
...
+
+1. We propose to stick to the assumption that ST can change for every sample and don't try to prematurely optimize for the special cumulative best case. Rationales:
+
+* This unlocks the most amount of benefits (e.g. also delta) for the same amount of work, it makes code simpler.
+* We don't know if we need special cumulative best case optimization (yet); also, it would be for some "best" cases. Once we know we can always add those optimizations.
+
+2. Similarly, we propose to not have new special ST storage cases per metric types beyond the existing difference between native histograms and float metrics. Some systems allow optional STs on gauges (e.g. OpenTelemetry). We propose keeping that storage flexibility.
+
+3. We propose to treat ST as an *optional* data everywhere, simply because it's new and because certain metric type do not need it (gauges). For efficiency and consistency with scrape and Remote Write protocols we treat default value `int64(0)` as a "not provided" sentinel. This has a consequence of inability to provide the ST of an exact 0 value (unlikely needed and if needed clients needs to use 1m after or before.
```
krajorama: nit: I've always heard people say 1 millisecond before or after , not 1 minute.

```suggestion
3. We propose to treat ST as an *optional* data everywhere, simply because it's new and because certain metric type do not need it (gauges). For efficiency and consistency with scrape and Remote Write protocols we treat default value `int64(0)` as a "not provided" sentinel. This has a consequence of inability to provide the ST of an exact 0 value (unlikely needed and if needed clients needs to use 1ms after or before.
```

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,420 @@
...
+
+4. Let's go through all the areas that would need to change:
+
+### Feature flag `st-storage`
+
+To develop and give experimental state to users, we propose to add a new feature flag `st-storage`. Similar to [exemplar-storage](https://prometheus.io/docs/prometheus/latest/feature_flags/#exemplars-storage) it will tell Prometheus to ingest ST and use new (potentially breaking compatibility) storage formats.
+
+We propose to have a single flag for both WAL, Block storage, etc., to avoid tricky configuration.
+
+Notably, given persistence of this feature, similar to example storage, if users enabled and then disabled this feature, users will might be able to access their STs through all already persistent pieces (e.g. WAL).
```
krajorama: nit:

```suggestion
Notably, given persistence of this feature, similar to example storage, if users enabled and then disabled this feature, users might be able to access their STs through all already persistent pieces (e.g. WAL).
```

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,420 @@
...
+
+This feature could be considered to be switched to opt-out, only after it's finished (this proposal is fully implemented) stable, provably adopted and when the previous LTS Prometheus version is compatible with this feature.
+
+#### Interaction with `created-timestamps-zero-injection`
+
+The existing zero-injection feature is useful and has been around long enough that some large users are relying on it. The new ST support will be in addition to or an alternative to zero-injection. In order to better differentiate zero-injection from st-per-sample, we propose to rename `created-timestamps-zero-injection` to `start-time-injection-at-scrape`.
```
krajorama: Let's not use "or". I propose to define it as an addition feature that does its thing after the zero-injection. This is consistent with the sentence about precedence later "In the case where a series has both scrape-injected zeroes and st-per-sample, the scrape-injected zeros take precedence.".

I'm not sure about the rename, even though maybe I brought it up in the meeting, let's just move the rename question to a normal issue in Prometheus. See also discussion in this [comment](https://github.com/prometheus/proposals/pull/87/changes#r3501637629).

```suggestion
The existing zero-injection feature is useful and has been around long enough that some large users are relying on it. The new ST support will be an additional feature situated in the ingestion pipeline *after* the zero-injection feature.
```

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,420 @@
...
+
+We also propose that the st setting be be tri-state on a per-series basis. Start Time per Sample support can be either `off`, `scrape`, or `per-sample`. Large installations may have a mix of implementations and there is no compatibility concern with having series have differing settings.
```
krajorama: Migrations are hard, I'm not sure it helps that much to have a per-series setting. I'd keep it to on/off in  global scrape config and per scrape config. 

Also this question is not related to the section it's in, let's add a "Configuration" section *before* this one that defines what configuration options we want. And we can then define how this section impacts the configuration if at all.

Related to [comment](https://github.com/prometheus/proposals/pull/87/changes#r3501652436).

proposals/0060-st-storage.md
```diff
@@ -0,0 +1,420 @@
...
+
+In the case where a series has both scrape-injected zeroes and st-per-sample, the scrape-injected zeros take precedence.
```
krajorama: To keep these two features separate, my idea of precedence would be to set ST to 0 when zero-inject is enabled at scrape time. So that effectively the ST feature doesn't do anything afterwards.

proposals/0060-st-storage.md
krajorama: @ywwg We've implemented opaque storage of ST and the above algorithm shared by Vilius for deciding between delta and cumulative with unknown start time: https://github.com/prometheus/prometheus/blob/06acd07312d4ae77ce0688624ef9c3044cad193f/promql/functions.go#L775

```suggestion
The stored ST is to be evaluated in the PromQL engine when the feature is enabled. We propose to utilize the ST for the purposes of counter reset detection and extrapolation at the beginning of time ranges.

The current reset detection takes into account only the value of two consecutive samples (`V[i-1]`, `V[i]`). We propose to take into account the start time of both samples as well. If the start time of the seconds sample is valid (`ST[i]!=0` and `ST[i]<T[i]`) and either its start time is after the previous sample (`T[i-1]<ST[i]`) or it is the same (`T[i-1]==ST[i]`), but the previous sample has a valid timestamp (`ST[i-1]!=0` and `ST[i-1]<T[i]`) then we detect a reset.

Note that this reset algorithm distinguishes between the case when a cumulative series starts with unknown timestamp (`ST[i-1]==T[i-1]`) or this is a delta series starting.

Extrapolation uses a known, in-range `ST[0]` to precisely detect the zero point of a series, instead of extrapolating.
```

krajorama (changes requested):
Proposing some changes.

proposals/0060-st-storage.md
bwplotka: Yea, I think we should ensure options in scrape config follow "ST" but given ff flag should go away with eventual stability I'd propose keeping it -- could be separate discussion, yes. 👍🏽

proposals/0060-st-storage.md
bwplotka: What's wrong with having both? Is it because ST=T on 0(fake) @ [ST, ST], right?

proposals/0060-st-storage.md
bwplotka: Added Config section.

proposals/0060-st-storage.md
bwplotka: Done

proposals/0060-st-storage.md
bwplotka: Mentioned that in the text.

krajorama: Yes, `ST[i]==T[i]` for the zero sample is one aspect, the other is the explicit ask from stakeholder (@SuperQ ) to not cause overhead when they choose to do zero inject instead.

krajorama: The wording isn't explicit about this , but @bwplotka you've resolved my comment. Do we want to state it or defer and keep somewhat vague?

krajorama: Discussed in chat. Since the options will be scrape level , you can totally just turn off st on the scrape where you want only zero injection.

krajorama (approved):
LGTM

ywwg (approved):

ywwg:
thanks for clarifying the zero-injection updates

PR #71, opened 2025-12-08
Entity Support in Prometheus

ArthurSens:


proposals/0071-Entity/01-context.md
```diff
@@ -0,0 +1,455 @@
...
+
+In OpenTelemetry, an entity is an object of interest that produces telemetry data. Entities represent things like services, hosts, containers, or Kubernetes pods. Each entity has a type (e.g., `k8s.pod`, `host`, `service`) and a set of attributes that describe it.
+
+This proposal adopts the Entity concept as the native Prometheus representation for what was previously expressed through Info metric conventions.
+
+#### Resource Attributes
+
+In OpenTelemetry, **resource attributes** are key-value pairs that describe the entity producing telemetry. These attributes are attached to all telemetry (metrics, logs, traces) from that entity. When OTel metrics are exported to Prometheus, resource attributes typically become labels on a `target_info` metric.
+
+#### Identifying Labels
```
krajorama: I think you mean attributes here, instead of labels. As this is the input.

proposals/0071-Entity/02-exposition-formats.md
```diff
@@ -0,0 +1,588 @@
...
+|---------|--------|-------------|
+| Entity type declaration | `# ENTITY_TYPE <type>` | Declares an entity type for subsequent entities |
+| Identifying labels | `# ENTITY_IDENTIFYING <label1> <label2> ...` | Lists which labels form the identity |
+| Entity instance | `<type>{<labels>}` | An entity instance (no value) |
+
+### Complete Example
+
+```
+# ENTITY_TYPE k8s.pod
+# ENTITY_IDENTIFYING k8s.namespace.name k8s.pod.uid
```
dashpole: In terms of the text format, this line seems to be the primary difference between this proposal and what an info metric currently provides.  Did you consider adding identifying labels to info metrics as an alternative?  That would address one of the current drawbacks of the new `info` function: https://prometheus.io/blog/2025/12/16/introducing-info-function/#future-development.

dashpole (commented):
I would love to see an alternatives considered section.

proposals/0071-Entity/02-exposition-formats.md
beorn7: Indeed. I think it is quite confusing if we create something that looks entirely new at first glance but is just a slight variation of something we had for a long time. I would rather keep the "info" nomenclature and extend the current info metric with the few missing pieces.

Also note that the `info` function could very well be the most natural way to "elevate" attributes from an entity into labels within PromQL (so that you can label-match/filter/aggregate on them). So it would be another benefit of keeping the info notion around.

ArthurSens:
> In terms of the text format, this line seems to be the primary difference between this proposal and what an info metric currently provides.  Did you consider adding identifying labels to info metrics as an alternative?  That would address one of the current drawbacks of the new `info` function: https://prometheus.io/blog/2025/12/16/introducing-info-function/#future-development.
> 
> _Originally posted by @dashpole in https://github.com/prometheus/proposals/pull/71#discussion_r2623824752_

> Indeed. I think it is quite confusing if we create something that looks entirely new at first glance but is just a slight variation of something we had for a long time. I would rather keep the "info" nomenclature and extend the current info metric with the few missing pieces.
> 
> _Originally posted by @beorn7 in https://github.com/prometheus/proposals/pull/71#discussion_r2624581564_

> The info metric concept (as it has existed ~forever) is essentially an implementation of entities. While it has the known drawbacks, it neatly fits into the existing simple Prometheus data and execution model.
> 
> _Originally posted by @beorn7 in https://github.com/prometheus/proposals/issues/71#issuecomment-3662184491_

> In OpenMetrics, info metrics are "native info metrics" and not gauges.
> In different news, I find "fundamentally flawed" needlessly drastic. It doesn't help the understanding to pretend that an entity doesn't share anything with a time series.
> 
> _Originally posted by @beorn7 in https://github.com/prometheus/proposals/pull/71#discussion_r2624605793_
            
I'm not against changing Info metrics slightly to get them closer to Entities! I'd actually love that since we can easily onboard Prometheus power users into this concept. They know the lingo already and what kind of things Info metrics can do for them, we would just make the experience better :) 

My two cents here is that this is an even more invasive change, we would be changing something that exists today and might have an impact on existing queries, dashboards, alerts, etc that already depend on existing Info-like metrics. Introducing a new concept would let users keep what they have unchanged, all existing queries would continue to work as they are today since we would be doing an additive change instead of a transformative one.

That said, if consensus from the community is to modify the existing Info metrics, I'll be happy to join you all and advocate for that too.

> Also note that the info function could very well be the most natural way to "elevate" attributes from an entity into labels within PromQL (so that you can label-match/filter/aggregate on them). So it would be another benefit of keeping the info notion around.

I'd love if we could simplify even more this part. If we could, at the same time, filter/aggregate on top of resource attributes with the same syntax we have for labels while also being able to present them separately in the query result, that would be my wet dream. While the info function does help when compared to traditional joins, not needing the info function at all is an even easier solution for the end-user. Have you had the chance to read the Querying proposal? Am I tripping too much there, or is that realistic?

> I would even make the opposite point and claim that an entity is very much a time series: An entity has a start and an end in time. Its descriptive labels are its data. They can even change over time. The data in form of descriptive labels are transferred at certain points in time, so it is fundamentally sampled.
> So the info metric (as defined by OpenMetrics, not as implemented in detail in current Prometheus) is a very good match for an entity. And it is totally a time series with data labels as sample values.
>
> _Originally posted by @beorn7 in https://github.com/prometheus/proposals/pull/71#discussion_r2624605793_

I totally agree with you here, I believe we have similar conclusions on how descriptive attributes behave and the proposal covers that aspect.

> IMO this is the proposed solution, but is listed as a goal. I would love to see this in terms of desirable user journeys, if possible.
> 
> _Originally posted by @dashpole in https://github.com/prometheus/proposals/pull/71#discussion_r2623837885_

Fair point! I'm not great at defining user journeys, but I'm getting some help from @amy-super and @vampirarte to get them defined :) Ana has done interviews with quite a few people (@beorn7, @juliusv, @jsuereth and @jpkrohling) and is designing wireframes and prototypes based on those interviews.

Ana has been sharing her work in Slack (feedback welcome to continue evolving):
* Wireframe for the table view of a query result[[1](https://cloud-native.slack.com/archives/C01AUBA4PFE/p1764695720803489)]
* Wireframe for using Resource Attributes to modify query results[[2](https://www.loom.com/share/66f3cc919bc94f08826dfc576e07a3ef)]

She's also working on a queryless experience that OTel users have been nudging us so much. She hasn't shared a video yet, but I think we can get an idea just based on a screenshot :) 

<details><summary>Queryless Entity explorer</summary>

image

</details>

On that note, an "Entity explorer" does sound a bit better than an "Info Explorer". I know folks have been big fans of info, but I don't know... sounds like an awkward naming to what they represent 😅 

> I would love to see an alternatives considered section.
> 
> _Originally posted by @dashpole in https://github.com/prometheus/proposals/pull/71#pullrequestreview-3583800387_

I explicitly made the decision to not add alternatives for two reasons:
* This proposal is already 6k lines long and it touches almost all parts of Prometheus. If I had alternatives included for all parts, it's hard to connect everything and the proposal would probably reach close to 10k lines.
* I see this proposal more as a live document 🙂 I'm not strongly opinionated on anything that I wrote, I just took a week to explore ideas and make a concise proposal. From now on I'd love to collect feedback from domain experts and let's evolve the proposal together. If our discussions show that parts of the proposal needs to change, I'm happy to apply those changes.

proposals/0071-Entity/06-querying.md
```diff
@@ -0,0 +1,555 @@
...
+# Regex matching on descriptive labels
+http_requests_total{service.version=~"2\\..*"}
+```
+
+**Query Execution for Descriptive Label Filters:**
+
+1. Select all series that might match (based on metric name and any indexed labels)
+2. For each series, look up correlated entities
+3. Get descriptive labels at evaluation timestamp
+4. Apply the filter: keep series where enriched labels match
```
beorn7: It should be noted that this is not very efficient in this naive form.

I would assume that a classical join query is probably already more efficient now. But in whatever way things are implemented, this can be done more efficiently, at least behind the scenes. Spelling out the least efficient way of doing it as if it were a satisfactory solution is misleading. This should at the very least mention that this can (and most likely) has to be improved.

proposals/0071-Entity/06-querying.md
```diff
@@ -0,0 +1,555 @@
...
+**Important:** Descriptive label changes do NOT create new series. The series identity is defined by:
+- Metric name
+- Original metric labels
+- Entity identifying labels (correlation keys)
+
+Descriptive labels are metadata that "rides along" with samples, not part of series identity.
+
+---
+
+## The Entity Type Filter Operator
```
beorn7: Ooooof, a whole new mini language for just one thing.

Is there really no ways this can be modeled with existing label matchers? Put the entity type into a label `entity_type`? Or maybe `__entity_type__` to not make it collide with a normal label that happens to have that name…

proposals/0071-Entity/01-context.md
```diff
@@ -0,0 +1,455 @@
...
+- Are required to identify which entity produced the telemetry
+
+Examples:
+- `k8s.pod.uid` or (`k8s.pod.name`,`k8s.namespace.name`) for a Kubernetes pod
+- `host.id` for a host
+- `service.instance.id` for a service instance
+
+#### Descriptive Labels
+
+**Descriptive labels** provide additional context about an entity but do not serve to uniquely identify it. These labels:
```
aknuds1: As described regarding "Identifying Labels", I fear that we are here simply duplicating OTel Entity Data Model design. It feels unnecessary.

proposals/0071-Entity/01-context.md
```diff
@@ -0,0 +1,455 @@
...
+
+---
+
+## Goals
+
+This proposal aims to achieve the following:
+
+### 1. Define Entity as a Native Concept
+
+Prometheus should recognize Entities as a distinct concept with their own semantics, separate from metrics. Entities represent the things that produce telemetry, not the telemetry itself.
```
aknuds1: How is this a better approach than the existing idea of modeling Resources and Entities as [persisted Prometheus metadata](https://docs.google.com/document/d/1epBslSSwRO2do4armx40fruStJy_PS6thROnPeDifz8/edit?tab=t.0#heading=h.5sybau7waq2q)?

proposals/0071-Entity/01-context.md
```diff
@@ -0,0 +1,455 @@
...
+
+### 2. Support Identifying and Descriptive Label Semantics
+
+Entities should allow declaring which labels are identifying (forming the entity's identity) and which are descriptive (providing additional context that may change over time).
+
+### 3. Improve Query Ergonomics
+
+Reduce or eliminate the need for `group_left` when attaching entity labels to related metrics. The common case should be simple.
```
aknuds1: Why is the `info` function not mentioned?

proposals/0071-Entity/01-context.md
```diff
@@ -0,0 +1,455 @@
...
+
+### 4. Optimize Storage for Metadata
+
+Entities store string labels and change infrequently. Storage and ingestion should be optimized for this pattern, rather than treating them as time series with constant values.
```
aknuds1: The title says "metadata", but then moves on to refer to "entities". Why should we not model metadata more generally, and treat entities as a subset of this?

aknuds1 (commented):
My take is the proposal's premise needs to change radically because:
1. The idea of introducing native support for Prometheus entities duplicates/conflicts with the ongoing work for [adding persisted metadata](https://github.com/prometheus/prometheus/issues/12608) to Prometheus.
1. The proposal presents info metrics as a problem. As I see it, this an objective misrepresentation, because info metrics are instead the _current_ solution to Prometheus' need for persisting metadata. While they are not a perfect solution, we already envision for persisted metadata to be the upcoming and improved solution.

PR #76, opened 2026-03-10
Proposal for memory limiting in the scrape loop

dashpole:
As @bwplotka presented in his "Scrape Trolley Dillema" talk last year at promcon, Prometheus could use a mechanism to prevent OOMs caused by short-term memory usage of scraping targets. This has also been a request from some users of the OTel collector's Prometheus receiver.

PoC: https://github.com/prometheus/prometheus/compare/main...dashpole:prometheus:memory_limiter_simple

cc @bernot-dev @ArthurSens


bwplotka (commented):
IMO this is a great plan, something I'd like to see 👍🏽

proposals/0076-scrape-memory-limiter.md
```diff
@@ -0,0 +1,126 @@
...
+
+If this flag is absent, the memory limiter will not be active and the configuration block will be ignored, even if configured in the prometheus configuration.
+
+### Debuggability and User Experience
+
+Understanding that data is missing and *why* it is missing is a critical part of the user experience. This feature caters to two personas:
+
+**1. The Application Owner:**
+Application owners need to understand why their specific application failed to be scraped.
+* **Up Metric:** The `up` metric for their dropped target will record a `0`. This is the standard mechanism to indicate a failed scrape, which preserves their existing alerts on the `up` metric.
```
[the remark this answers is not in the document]
bwplotka: Agree, let's keep proposal simple and create an issue - I was thinking about same thing - how to capture other reasons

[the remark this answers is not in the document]
bwplotka: added https://github.com/prometheus/prometheus/issues/18284

bwplotka (approved):
I'm supportive. It should break the ice on the non trivial issues you listed and it's elegant. This would also help with Otel-collector Prometheus scrapes.

 * https://github.com/prometheus/prometheus/issues/17109
 * https://github.com/prometheus/prometheus/issues/13939
 * https://github.com/prometheus/prometheus/issues/11306
 * https://github.com/prometheus/prometheus/issues/16917


Disclaimer @dashpole works with me at Google, so I'm not going to merge until we have buy-in from other maintainers.

WDYT @krajorama @roidelapluie @bboreham @ArthurSens ?

saswatamcode (approved):
Would love to see this!

proposals/0076-scrape-memory-limiter.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Fairness Mechanisms
+
+The initial implementation of the memory limiter proposed above might inadvertently starve small, critical targets when a noisy neighbor introduces memory pressure. Future iterations could introduce scheduling algorithms to ensure fairness. Advanced approaches like [Deficit Round Robin (DRR)](https://en.wikipedia.org/wiki/Deficit_round_robin) can mathematically guarantee fairness across targets during memory pressure, isolating the disruption to high-cardinality targets.
+To implement fairness, the mechanism will need to predict the cost of a scrape. This prediction should be based on the **total number of samples** from the target's previous scrape, *not* the number of *new series* added. New series are highly volatile (a target rotating a label will add many new series in one scrape, but zero in the next), making them a poor heuristic for proactive load shedding. Total samples accurately correlate with the short-lived parsing overhead the scrape loop will incur.
```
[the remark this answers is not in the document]
dashpole: We don't actually need to predict the memory cost in MB (I can update this). That will change over time as we change the server code. We just need to be able to:

* Estimate the **relative** short-term cost of different targets so that we throttle targets relative to the costs they impose on the server.
* Throttle more aggressively (or less aggressively) depending on how far above or below the limit we are.

The important point is just that the number of samples (not number of new samples) is probably the best thing to compare to determine the cost of targets.  We might want to weight that by scrape interval (i.e. the average number of samples per second is the actual metric) to avoid incentivizing shorter scrape intervals as a way to get more priority when memory limited.

dashpole:
Discussed this at the Prometheus Dev-summit today.  Feedback:

* @SuperQ has a lot of ideas around this, and we should discuss further.
* @bboreham says there is something similar in Mimir (but is AGPL), so this general idea has some backing by production usage.
* Memory usage is cyclic, and tends to be highest right before compaction.
* Series already in the index don't add to long-term memory usage.  It is new series that grow the long-term memory usage.

SuperQ:
Another memory mitigation that could be done would be to pause rule evaluations. Specifically recordings could be paused since they run queries which can be an alloc churn impact.

dashpole:
@SuperQ Do you think it should just pause rule evaluations, or should it cause all PromQL queries to fail?  If the issue is queries, seems like pausing all queries would be better?  OTOH, intermittently not being able to query prometheus seems like it would make it hard to tell what is going wrong.

SuperQ:
That's a good question. Pausing rules is nice because you can, in theory, resume without loss.

Pausing _all_ PromQL is also a good option.

For "hard to tell what's going on" is why meta-monitoring exists. Having a separate Prometheus-to-watch-Prometheus is usually recommended for that. :grin:

yeya24 (commented):
I like this proposal and we saw quite a lot of OOM kills of scraping due to quick churning during deployment or any sudden load increase.

One thing I am wondering is that if there is a way to deal with scraper OOM kills without losing data. That's definitely more for future but if the scraper can scrape the metric snapshot with a timestamp then the missing metrics can be backfilled with OOO

proposals/0076-scrape-memory-limiter.md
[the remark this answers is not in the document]
yeya24: > How we could estimate how much memory one scrape will take

Not memory but it would be really nice if we extend the `/metrics` endpoint for `HEAD` so that we can somehow detect the cardinality/samples of the endpoint. This can help with preventing OOM kills

SuperQ:
It would be interesting to have some heuristic that uses `scrape_series_added`, or maybe something like this:

```
increase(prometheus_tsdb_head_series_created_total[5m)
/
prometheus_tsdb_head_series
```

Canceling scrapes is, IMO, an action of last resort because it means we're dropping data permanently.

dashpole:
I've had some time to think about how to incorporate recording rules and queries. Taking a broader look at the entire prometheus server, I think we could apply memory limiting in some form to:

* `scrape_configs` (this proposal)
* `otlp`: Reject incoming OTLP metrics
* `remote_read`: Reject incoming remote_read 
* `rule_files`: Pause recording rules (I assume we don't want to pause alerting rules!?)
* compaction: Pause TSDB compaction.
* Parts of the API (queries, targets, series, metadata)?

I think the best way to address this is to make the feature naming more generic ("memory_limiter" instead of "scrape_memory_limiter") and to bundle memory limiting behavior behind a "hard" limit (data loss), and a "soft" limit (no data loss). That seems like it would be easy to understand, comprehensive, and configurable enough for most usage.

When the "soft" limit is met:

* Pause compaction
* Pause recording rules
* Reject PromQL queries (and other API queries?)

When the "hard" limit is met:

* Fail scrapes
* Reject OTLP
* Reject remote read

Question for reviewers:

Would you prefer I:

1. Expand the scope of the proposal to include all of the above behavior?
2. Just change the name of "scrape_memory_limiter" to "memory_limiter", and leave the rest of the topics (soft limits, recording rules, compaction, etc) as future enhancements?
3. Expand the scope of the proposal to only include pausing recording rules (in addition to failing scrapes).

SuperQ:
I think expanding the proposal to support multiple mitigation options is a good idea.

Since all of the options discussed so far have pretty impactful downsides different users are going to want opt-in to different options.

We likely want to support configuring different options, have some kind of tunable heuristics, etc.

But all of it is going to need to be tested for various failure modes.

dashpole:
@SuperQ I expanded the proposal. The things I want to make sure we get right during the proposal stage are:

* Categories: Did I miss any important sources of short-term memory usage? I've initially excluded standard promql queries, or other debugging parts of the API surface (e.g. targets or metadata pages).
* Soft and hard limits: Does this categorization of mitigations into "destructive" (hard) vs "non-destructive" (soft) make sense?  Or should we allow setting more granular memory thresholds for each mitigation?
* Configuration: Is turning on/off each mitigation enough configurability?  Or will we need something more granular at the global level?

proposals/0076-memory-limiter.md
```diff
@@ -0,0 +1,218 @@
+# Memory Limiter
```
dashpole: The mechanism does not bound peak memory for the failure mode it exists to prevent. add a subsection under §How stating (a) the exact placement of the check (before `scraper.scrape`, before `readResponse`, or before `app.append`), (b) what the resulting high-water bound is in terms of in-flight scrapes, and (c) either a per-scrape byte budget (see F4) or an explicit Non-Goal: "this proposal bounds sustained intake, not the peak allocation of any individual scrape; use `bodysizelimit` for that." If a global in-flight-scrape cap is the intended backstop, propose it here — it is a small change and it is the piece that turns "unbounded sum" into "bounded sum".

dashpole: No hysteresis, no dwell time, no bound on mitigation duration — and indefinite compaction pause is a positive feedback loop. specify a control policy, not just thresholds: separate `engageratio` / `disengageratio` (or a single ratio plus a hysteresis band), a minimum dwell time in each state, and a maximum continuous engagement duration after which compaction resumes regardless (with a log line and a metric).

dashpole: Capacity-planning math uses the wrong `GOGC` default, and the default ratios put healthy servers permanently in the soft-limit state. correct `GOGC=100` → `75` and `--runtime.gogc` → the `runtime.gogc` config field throughout; redo the 35% figure as 40%; and either raise `softlimitratio` so the clean band covers realistic servers, or derive the soft threshold from `GOGC` at runtime (`soft ≈ min(configured, k / (1 + GOGC/100))`) rather than hard-coding a ratio that silently depends on a parameter the operator can change.

dashpole: `bodysizelimit` and `sample_limit` are not reconciled — the most obvious reviewer objection is unaddressed. add a subsection `Relationship to existing scrape limits` covering `bodysizelimit`, `samplelimit`, `targetlimit`, `labellimit`, and add at least two entries to §Alternatives: (a) non-zero default `bodysizelimit`, (b) pressure-driven dynamic `bodysizelimit` / `samplelimit`.

dashpole: Skip-scrape failure semantics: the staleness-marker justification overstates the cost, and the cache-flush hazard is missed. rewrite the parenthetical in L78 to state the real cost (one interval's markers, once) and pick the trade-off on that basis; add an explicit requirement that a skipped scrape must not flush the scrape cache (call `iterDone(false)`-equivalent or bypass the cache path entirely) and add that to the §How We Test list; document the `up == 0`-without-staleness semantics and the fleet-wide alert consequence in §Debuggability.

dashpole: `/gc/limiter/last-enabled:gc-cycle` is a cycle number, not a boolean — the hard-limit trigger is underspecified. specify the comparison and the recency window, e.g.

dashpole: The signal measures the Go runtime's footprint, not the process's — mmap'd TSDB memory is excluded by definition. one paragraph in `Relationship to Go Runtime Parameters` stating what `pressure_ratio` does and does not include, why the residual (mmap'd chunks, cgo, binary mappings) is expected to be covered by the `auto-gomemlimit` reserve, and — worth considering — a startup warning if `GOMEMLIMIT ≥` the detected container limit while the memory limiter is enabled.

dashpole: "In-use memory responds immediately when load is shed" is asserted without evidence, and is the same class of lag Alternative 4 was rejected for. run the equivalent experiment (burst → shed → measure `mappedReady` decay to baseline) and put the number in the document, the way 0055 puts its Prombench/Reddit findings in `Experimental Analysis and Trade-offs`.

dashpole: The problem statement is entirely unquantified. add an `Experimental Analysis` subsection with a heap-profile breakdown of a large scrape, an estimate of memory reclaimed per mitigation, and the burst→OOM timeline that motivates `check_interval: 100ms`.

dashpole: Refusing to start when `GOMEMLIMIT` is unset turns a warning into a crash-loop. downgrade to a startup `Error`-level log plus an explicitly inert state that is visible in metrics and on `/status`; reserve the hard failure for the case where the operator has explicitly configured `memory_limiter` in YAML and `GOMEMLIMIT` is unset, if a hard failure is wanted at all.

dashpole: Internal inconsistency: the test plan references metrics and a state model the document does not define. pick one.

dashpole: Remote-read rejection has the same external-dependency problem the document calls out for recording rules. extend the operator guidance paragraph to cover `rejectremoteread` and `reject_federation`, or note explicitly why 503+retry makes them materially safer than a skipped rule evaluation.

dashpole: House-style deviations. - Heading level.

dashpole: The standing approvals predate the substantive rework. re-request review from the four maintainers bwplotka named, and add SuperQ's churn-rate heuristic to §Alternatives with a rationale for why a memory-pressure signal was preferred over a churn signal.

proposals/0076-memory-limiter.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Audience
+
+Prometheus operators running in memory-constrained environments who need to protect the server from unpredictable memory spikes from various sources.
+
+## Non-Goals
+
+- Fairness and per-job QoS controls are out of scope for the initial implementation.
+- This does not address long-term memory leaks. It is designed to handle spikes and overload scenarios.
+- This proposal bounds sustained global memory intake across subsystems rather than bounding the peak allocation of any individual scrape. Per-target peak burst bounds remain the responsibility of existing controls like `body_size_limit`.
```
dashpole: The Non-Goal delegates per-scrape peak bounds to `body_size_limit`, but that defaults to `0` (unlimited) — `docs/configuration/configuration.md:106` — so on a default Prometheus nothing bounds the peak. Worth stating that here. Also, `sample_limit` can't do what L121 claims: it's enforced in the parse loop after `readResponse` has already buffered the whole decompressed body, so it bounds ingested samples, not response allocation.

proposals/0076-memory-limiter.md
```diff
@@ -0,0 +1,235 @@
...
+
+Periodically (default `check_interval: 100ms`, consuming ~0.001% CPU at 10 Hz), a background routine calculates the memory pressure ratio:
+`pressure_ratio = in_use_memory / GOMEMLIMIT`
+
+The limiter maintains two state thresholds:
+* **Soft Limit**: Reached when `pressure_ratio >= soft_limit_ratio` (default `0.70`).
+* **Hard Limit**: Reached when `pressure_ratio >= hard_limit_ratio` (default `0.85`), or immediately if Go's runtime GC CPU limiter has been active recently. Because `/gc/limiter/last-enabled:gc-cycle` is a `uint64` cycle counter rather than a boolean, the limiter detects engagement by checking if `last_enabled_gc_cycle >= current_completed_gc_cycles - 1` (queried alongside `/gc/cycles/total:gc-cycles`). These default ratios provide a balanced safety margin: 70% triggers early, non-destructive load shedding, while 85% leaves enough remaining heap headroom for Go's garbage collector to reclaim transient memory before hitting an OOM crash.
```
dashpole: `0` is the sentinel for "never enabled" (`runtime/metrics/description.go:288`: *"The first GC cycle is cycle 1, so a value of 0 indicates that it was never enabled"*), and this predicate doesn't exclude it — on a server that has never engaged the limiter, `0 >= 1 - 1` is true from the moment the first GC cycle completes until the second one does, asserting the hard limit at ~5% pressure. Worth noting too that `total - 1` underflows in `uint64`, so the natural rewrite `last_enabled + 1 >= total` widens the window to cover process start. Suggest `last_enabled != 0 && last_enabled + 2 >= total_cycles`.

proposals/0076-memory-limiter.md
```diff
@@ -0,0 +1,235 @@
...
+
+**At Hard Limit (Discard work to prevent crashes):**
+- **Fail Scrapes**: Skip scrapes to prevent allocation of memory for new samples. The check is performed at the start of `scrapeLoop.scrape()` before making the HTTP fetch and allocating response decoding buffers. To avoid causing a synchronized WAL append storm when memory is exhausted, skipped scrapes bypass appending per-series staleness markers, letting values carry forward under the standard 5-minute lookback. Crucially, a skipped scrape **must preserve the existing `scrapeCache`** (bypassing cache flush/eviction paths); clearing the cache on a skipped scrape would trigger a severe re-allocation storm when scraping resumes upon recovery, defeating the limiter's purpose.
```
dashpole: `scrapeLoop.scrape()` doesn't exist — it's `scrapeAndReport` on `scrapeLoop` (`scrape/scrape.go:1488`) and `scrape(ctx)` on `targetScraper` (`:824`). The distinction matters: at the very top of `scrapeAndReport` the check returns before the `sl.report` defer, so no `up = 0` is written; at the forced-error branch (`:1545`) `up = 0` works but the next line calls `app.append([]byte{}, …)`, which is the staleness-marker walk this paragraph says to skip — and so does the generic scrape-failure path (*"A failed scrape is the same as an empty scrape, we still call sl.append to trigger stale markers"*). Worth naming the exact position, since none of the existing hooks gives you both `up = 0` and no staleness markers.

jmacd:
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45439#issuecomment-5684820227

proposals/0076-memory-limiter.md
```diff
@@ -0,0 +1,235 @@
...
+
+**At Hard Limit (Discard work to prevent crashes):**
+- **Fail Scrapes**: Skip scrapes to prevent allocation of memory for new samples. Specifically, the check is performed within `scrapeLoop.scrapeAndReport` immediately after establishing the deferred `sl.report` hook (ensuring target health and `up = 0` are recorded with an explicit scrape error like `errMemoryLimitExceeded`), but before calling `targetScraper.scrape(ctx)` to prevent HTTP fetch and response buffer allocations. Unlike standard scrape failure paths—which invoke `app.append([]byte{}, …)` / `sl.append` to walk existing series and append staleness markers—the memory limiter introduces an explicit short-circuit return that skips the staleness marker walk entirely to prevent a synchronized WAL append storm when memory is exhausted, letting values carry forward under the standard 5-minute lookback. Crucially, this skip path **must preserve the existing `scrapeCache`** (bypassing cache flush/eviction paths); clearing the cache on a skipped scrape would trigger a severe re-allocation storm when scraping resumes upon recovery, defeating the limiter's purpose.
```
bwplotka: Interesting point about cascading failure - I think it makes sense

proposals/0076-memory-limiter.md
```diff
@@ -0,0 +1,235 @@
...
+    # Fraction of GOMEMLIMIT (in-use memory) at which non-destructive mitigations engage.
+    soft_limit_ratio: 0.70
+
+    # Fraction of GOMEMLIMIT (in-use memory) at which destructive mitigations engage.
+    hard_limit_ratio: 0.85
+
+    # Granular controls to enable/disable specific mitigations
+    enforcement:
+      # Soft Limit
+      pause_block_compaction: true
```
bwplotka: Maybe overkill but mitigation pieces could have  memory ratios setting instead (float) so you can tune those up to you wish 🤔

bwplotka: then you don't need soft_limit_ratio and hard_limit_ratio even

bwplotka (approved):
I would love to try this out, left some comments, but curious what others think cc @SuperQ @GiedriusS

PR #77, opened 2026-03-18
proposal: Use Start Timestamp in `rate`-like functions for deltas

vpranckaitis:
_This proposal is currently WIP._

This proposal is a part of [PROM-48](https://github.com/prometheus/proposals/pull/48) work. It focuses on using start timestamps in `rate`-like functions to enable querying OTel delta temporality counters and histograms. Is is also closely related to [PROM-60](https://github.com/prometheus/proposals/pull/60), which describes TSDB support for start timestamps.


proposals/0077-use-start-timestamp-in-rate-like-functions.md
```diff
@@ -0,0 +1,247 @@
...
+                               ...  [ST3; T3]  // new sequence
+
+GCP cumulative:
+   [ST0; - T0]
+   [ST1; -------- T1]
+   [ST2; ----------------- T2]
+                               ...  [ST3; T3]  // new sequence
+```
+
+### Detecting ST counter resets between two successive datapoints
```
ywwg: This section provides a good reason for why the "free aggregation" approach of dropping labels and hoping for the best will not work. Maybe we can call that out explicitly here or at the end of the proposal?  We can say that when the engine encounters that sort of mashed-together series, the reset detection fails and the data is considered too messy to process.

proposals/0077-use-start-timestamp-in-rate-like-functions.md
```diff
@@ -0,0 +1,247 @@
...
+
+This proposal suggests looking at no more than two successive datapoints at a time for reset detection, and thus processing the datapoints inside the `rate`-like function window pair by pair. To tell whether there was a reset between two datapoints, it is enough to look at the start and datapoint timestamps of those two datapoints.
+
+One might argue that looking at more datapoints might provide more insight into the behavior of the timeseries that is being processed. However, it might also increase the likelihood of inconsistent results between query steps. For example, at some query step the `rate()` function might make a decision by judging in tandem datapoints at `T[1]`, `T[2]` and `T[3]`.
+However, the same `rate()` function will only see datapoints `T[1]` and `T[2]` at an earlier step, and only datapoints `T[2]` and `T[3]` at a later step. If this leads to a drastically different decision, the `rate()` function could produce non-uniform results across the steps.
+
+Nevertheless, more than two datapoints could be analyzed for producing info and warning messages about ST values (e.g. if they are invalid, or if there is a collision between two datapoint streams).
```
ywwg: let's have an explicit section on the warnings we think would be useful and plan to return. I think that would be a valuable part of this proposal to help guide implementation

PR #85, opened 2026-06-02
proposal: add info-metric label discovery APIs for info() autocomplete

aknuds1:
Editors completing `info(<expr>, {…})` need data-label names and values from info metrics scoped to the expression being edited. This proposal adds two dedicated Search API operations:

- `GET|POST /api/v1/search/info_labels` searches non-identifying data-label names.
- `GET|POST /api/v1/search/info_label_values` searches values for one exact `label`.

Both endpoints accept repeated full `data_match[]` matchers, including `__name__` matchers selecting info metric families, and an optional instant-vector `expr`. Expression-derived storage selection follows the lookback, `offset`, and `@` semantics of an instant `info()` evaluation.

The design builds on PROM-74, reusing the Search API’s storage interfaces, search, ranking, and NDJSON streaming infrastructure. Dedicated routes keep their info-specific scoping and expression-dependent time semantics explicit.

Both require `search-api` and `promql-experimental-functions`. A single per-request timeout covers expression evaluation and storage search. Requests containing `expr` require the same authorization as `/api/v1/query`.

The proposal defines result limits, expression-derived matcher bounds, fail-closed storage capability checks, and stream completion, warning, and truncation semantics. Suggestions are indexed candidates and do not guarantee a successful runtime join.

Work-in-progress implementation: [prometheus/prometheus#17930](https://github.com/prometheus/prometheus/pull/17930).

### DIFF
--- a/proposals/0085-info-labels-endpoint.md
+++ b/proposals/0085-info-labels-endpoint.md
@@ -1,4 +1,4 @@
-## Info-metric label discovery API for `info()` autocomplete
+## Info-metric label discovery APIs for `info()` autocomplete
 
 * **Owners:**
   * Arve Knudsen [@aknuds1](https://github.com/aknuds1) arve.knudsen@gmail.com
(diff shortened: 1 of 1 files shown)


proposals/0085-info-labels-endpoint.md
```diff
@@ -0,0 +1,278 @@
...
+| `limit`          | int ≥ 0                  | No       | 100, capped by `--web.search.max-limit` | Maximum number of label NAMES returned. Same `has_more` semantics as PROM-74.                                                                                                                                             |
+| `values_limit`   | int ≥ 0                  | No       | 0 (no cap)                              | Maximum number of values returned per label. New to this endpoint.                                                                                                                                                        |
+| `batch_size`     | int > 0                  | No       | 100                                     | As per PROM-74.                                                                                                                                                                                                           |
```
colega: What does a batch size limit in this API? Label names or label values?

proposals/0085-info-labels-endpoint.md
```diff
@@ -0,0 +1,278 @@
...
+
+This is a new gap created by the introduction of `info()` itself — it is not specific to any one editor or vendor. Anyone building a PromQL editor that supports `info()` will hit the same problem.
+
+The in-tree client implementation lives in `web/ui/module/lezer-promql/src/client/`; any PromQL editor that links it (Prometheus' own React UI via `codemirror-promql`, Grafana, others) inherits `/info_labels` autocomplete.
+
+### Pitfalls of the current solution
+
+Constructing the same response with the endpoints that exist today requires either fetching far more than needed or multiple round-trips per autocomplete suggestion.
+
+* **`/api/v1/labels` + per-name `/api/v1/label/{name}/values`** returns *all* labels matching a selector, including labels carried by the base metric, not just data labels on info metrics. Filtering client-side requires downloading the full universe of labels and then making one `/label/{name}/values` call per label of interest (N+1).
```
colega: This works for the existing autocomplete implementations for PromQL, why doesn't it work for the info function?

proposals/0085-info-labels-endpoint.md
```diff
@@ -0,0 +1,278 @@
...
+* `match[]` is not accepted. Scoping happens via `metric_match` (which info metrics to consider) and `expr` (which identifying-label values to restrict to). Accepting `match[]` as well would create two equivalent matcher mechanisms with unclear precedence.
+* The response unit is one record per data label NAME, carrying that label's values inline. Search-api emits names and values via separate endpoints; collapsing them here matches the autocomplete UI pattern: each label name is a top-level suggestion, and selecting one reveals its values.
+* The response is NDJSON (`application/x-ndjson`) with the same batch + trailer contract as PROM-74.
+* The endpoint is dual-gated: `--enable-feature=search-api` covers the NDJSON + parsing infrastructure it reuses; `--enable-feature=promql-experimental-functions` covers the only consumer (`info()`). Either missing flag returns the standard Prometheus JSON error with `errorType: unavailable` and a flag-specific message.
+
+### `GET|POST /api/v1/info_labels`
```
tcp13equals2: Given that this new endpoint leverages params and NDJSON from the search endpoint, perhaps this should be `/api/v1/search/info_labels`

proposals/0085-info-labels-endpoint.md
tcp13equals2: Note that I believe we added a max batch size ceiling. 10,000?

proposals/0085-info-labels-endpoint.md
```diff
@@ -0,0 +1,278 @@
...
+* **Whether to accept `match[]` in addition to `expr`.** The current design rejects it on cohesion grounds; reviewers may push back.
+* **Whether identifying labels should be configurable per request.** Today the set is `{"job", "instance"}` from `infohelper.DefaultIdentifyingLabels`. Per-request override is plausible if downstream implementations have different identifier conventions.
+* **One combined `info-labels-api` flag.** Rejected here on grounds of flag proliferation, but worth re-litigating if reviewers prefer a single-flag operator UX.
+
+## Alternatives
+
+### 1. Extend `/api/v1/search/label_names` to optionally return values per name
+
+Would collapse two endpoints into one. Rejected on cohesion grounds: PROM-74 keeps names and values in separate endpoints precisely so that each endpoint's response shape stays simple. The values payload is only useful when the names are scoped to info metrics, so the coupling does not belong on the general search endpoint.
```
tcp13equals2: I would not dismiss this option so quickly. Part of the reason for the search results format being objects/maps per record was so that additional data could be augmented with metric names, label names or label values.  

The idea of extending the response format is noted https://github.com/prometheus/proposals/blob/main/proposals/0074-new-labels-values-api.md#extensibility-for-mimir-thanos-cortex.

This endpoint could be extended to include a "include_values=true&values_limit=10" and the values collections could be decorated into each label record.

We already have the "only useful in a certain context" problem on the metric_names endpoint. It supports the include_metadata=true to decorate in metric metadata records for each metric name. But this only can decorate metric names where we have metadata. It's acceptable that this part of the response may not be there.

proposals/0085-info-labels-endpoint.md
aknuds1: I thought about this. The (subjective) reason for not going that route was that info function autocomplete endpoint is not in the same logical family. I'm open to rethinking this, however.

proposals/0085-info-labels-endpoint.md
aknuds1: Thanks for the feedback @tcp13equals2, will consider.

itsmylife:
I've implemented a [PoC](https://github.com/grafana/grafana-prometheus-datasource/pull/244) in `grafana-prometheus-datasource` using your [PoC](https://github.com/aknuds1/prometheus) in `arve/info-autocomplete` branch. 

From PoC stand point it's working nicely. But we should challenge the return value data type (instead of returning all potential labels/values we might return labels and then values for the selected label. In simple terms I'd like to follow current `labels` and and `label/<label-key>/values` style. I think this'll make auto complete easier. It'll also help client to fetch less data.

proposals/0085-info-labels-endpoint.md
aknuds1: Existing label search API doesn't work for `info` function autocomplete because it can't derive info series scope. I.e. the `info` function autocomplete API has to implement `info` semantics, in order to return data labels matching the expression.

aknuds1:
> But we should challenge the return value data type (instead of returning all potential labels/values we might return labels and then values for the selected label.

@itsmylife This dual-endpoint scheme has been implemented. Thanks for the feedback!

proposals/0085-info-labels-endpoint.md
aknuds1: In the original combined response format this was ambiguous. The proposal now splits the API; `batch_size` controls the number of result records in each NDJSON batch line: Label name records on `/api/v1/info_labels`, and label value records on `/api/v1/info_label_values`.

proposals/0085-info-labels-endpoint.md
aknuds1: After revising the proposal in the meantime, I kept these as top-level `info_*` endpoints. The reason being that the endpoints (now split into two) are specific to `info()`. Please let me know if you still think nesting them under `/api/v1/search/` would be preferable.

proposals/0085-info-labels-endpoint.md
aknuds1: Thanks! The current revision documents `batch_size` as a positive integer capped at 10,000.

proposals/0085-info-labels-endpoint.md
aknuds1: The proposal has been significantly revised in the meantime. Are you suggesting though to use `/api/v1/search/label_names` for `info` function autocomplete? I'm ruling that out for the reason that the autocomplete API needs `info()` specific semantics.

proposals/0085-info-labels-endpoint.md
```diff
@@ -0,0 +1,261 @@
...
+| `start`, `end`   | rfc3339 / unix timestamp      | No       | last 1h                  | Storage search window when `expr` is absent. With `expr`, `start` must parse but is ignored for range selection and ordering validation; `end` supplies the default `time`.                      |
+| `search[]`       | []string                      | No       |                          | At most 32 search terms, matched against names or values according to the endpoint. Multiple terms have OR semantics.                                                                            |
+| `fuzz_threshold` | int [0..100]                  | No       | 0                        | Fuzzy threshold, as in PROM-74.                                                                                                                                                                  |
+| `fuzz_alg`       | `jarowinkler` / `subsequence` | No       | `jarowinkler`            | Fuzzy algorithm, as in PROM-74.                                                                                                                                                                  |
```
tcp13equals2, line 75: subsequence is the default in the Prometheus implementation.

aknuds1: This should now be fixed - PTAL @tcp13equals2.

roidelapluie:
I would keep the endpoint behind /search. /search was meant partly for autocomplete, so it would be a fit. Also we could reuse some of the code and ensure some consistency between the endpoints.

I have no strong opinion about reusing endpoints with `scope=info` vs new endpoints. New enpoints might be easier to document and to use.

Note that we have also defined semantics that might be useful here, such as `include_medatata`.

aknuds1:
> I would keep the endpoint behind /search. /search was meant partly for autocomplete, so it would be a fit.

Thanks @roidelapluie - done!

> Note that we have also defined semantics that might be useful here, such as include_medatata.

Included a note on this.

roidelapluie:
Approved. I’d like it to remain experimental until we have a clearer picture of performance under realistic autocomplete usage. I suggest reusing the search-api toggle, with two separate entries in /api/v1/features for general search and info-label search, so clients can distinguish support for each.

roidelapluie (approved):

aknuds1:
Thanks @roidelapluie!

PR #86, opened 2026-06-27
proposal: Prometheus internal telemetry as an OTel semantic convention registry

nicolastakashi:
This proposal defines all metrics exported by the Prometheus binary as a formal OTel semantic convention registry. Making the schema machine-readable enables auto-generated instrumentation code, always-in-sync documentation, contract testing against live instances, and a lifecycle model for safe metric evolution across the Prometheus ecosystem.

Proof-of-concept implementation: https://github.com/prometheus/prometheus/pull/17868

### DIFF
--- a/proposals/0086-semconv-internal-telemetry.md
+++ b/proposals/0086-semconv-internal-telemetry.md
@@ -64,20 +64,21 @@ Prometheus maintainers and contributors. Operators and SREs who build dashboards
 Each package that exposes internal metrics has a `registry.yaml` file that fully describes every metric it owns. The exact layout (one file per package vs. a single file for the whole project) is an open question discussed below. The example below uses a per-package layout:
 
 ```yaml
-- id: metric.prometheus_tsdb_compaction_duration_seconds
-  type: metric
-  stability: stable
-  brief: Duration of compaction runs.
-  metric_name: prometheus_tsdb_compaction_duration_seconds
-  instrument: histogram
-  unit: s
-  annotations:
-    prometheus:
-      histogram_type: mixed_histogram
-      exponential_buckets: {start: 1, factor: 2, count: 14}
-      bucket_factor: 1.1
-      max_bucket_number: 100
-      min_reset_duration: "1h"
+groups:
+  - id: metric.prometheus_tsdb_compaction_duration_seconds
+    type: metric
+    stability: stable
+    brief: Duration of compaction runs.
+    metric_name: prometheus_tsdb_compaction_duration_seconds
+    instrument: histogram
+    unit: s
+    annotations:
+      prometheus:
+        histogram_type: mixed_histogram
+        exponential_buckets: {start: 1, factor: 2, count: 14}
+        bucket_factor: 1.1
+        max_bucket_number: 100
+        min_reset_duration: "1h"
 ```
 
 This file is the contract. Go code, documentation, and contract tests are all derived from it. The `annotations.prometheus` block carries Prometheus-specific details invisible to OTel (histogram variant, bucket configuration, callback-based gauges, labels fixed at construction time).
@@ -86,16 +87,16 @@ The repository structure is:
 
 ```
 pkg/semconv/
-  registry.yaml   ← source of truth, hand-authored
-  metrics.go      ← generated, DO NOT EDIT
-  README.md       ← generated, DO NOT EDIT
+  registry.yaml    ← source of truth, hand-authored
+  metrics.gen.go   ← generated, DO NOT EDIT
+  README.md        ← generated, DO NOT EDIT
 ```
 
 The Weaver templates and Rego policies used during generation are not committed to this repository. Where they live is an open question discussed below; the generation step resolves them at build time.
 
 ### Instrumentation code generation
 
-[OTel Weaver](https://github.com/open-telemetry/weaver) renders `registry.yaml` files into typed Go code via Jinja2 templates. A Makefile target regenerates all `semconv/metrics.go` files across the repository. CI enforces that generated files stay in sync with their registries.
+[OTel Weaver](https://github.com/open-telemetry/weaver) renders `registry.yaml` files into typed Go code via Jinja2 templates. A Makefile target regenerates all `semconv/metrics.gen.go` files across the repository. CI enforces that generated files stay in sync with their registries.
 
 A metric without labels generates a plain embedded type:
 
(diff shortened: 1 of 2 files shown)


roidelapluie:
I would like to assess what alternatives we have that could be built-in e.g. client_golang and would be lighter. And I would also want to better understand the scope of this: `a lifecycle model for safe metric evolution across the Prometheus ecosystem`.

nicolastakashi:
> I would like to assess what alternatives we have that could be built-in e.g. client_golang and would be lighter. And I would also want to better understand the scope of this: `a lifecycle model for safe metric evolution across the Prometheus ecosystem`.

@roidelapluie thanks for comment, I've updated a few things in the doc aiming to cover what you raised, let me know your thoughts

proposals/0086-semconv-internal-telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+
+Prometheus defines its own internal metrics as scattered `prometheus/client_golang` constructor calls across dozens of files. There is no machine-readable description of what Prometheus emits.
+
+This creates the following concrete problems:
+
+* Dashboard and alert authors who depend on these metrics have no canonical reference. They reverse-engineer what Prometheus emits by reading Go source or running a live instance.
+* There is no contract distinguishing stable public API metrics from internal implementation details. Metrics can be renamed, removed, or semantically changed without versioned signals.
+* Documentation, if it exists, is written separately from the code and drifts. There is no mechanism to keep it in sync.
+* There is no regression safety. Nothing prevents a code change from silently altering which metrics Prometheus emits or what labels they carry.
+* As Prometheus increasingly interoperates with OTel (OTLP ingestion, remote write, collectors), its own telemetry remains outside the OTel schema world and is invisible to tools that understand semantic conventions.
```
ArthurSens: I don't think that's a big problem, to be honest. It feels like you're tying the solution to the goals 🤔

proposals/0086-semconv-internal-telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+
+* Metric help strings are inconsistent: some describe counter semantics, others describe the event being counted.
+* Histogram bucket configurations are chosen ad hoc with no enforcement.
+* Units are absent from most metric definitions.
+* No lifecycle model exists: no way to mark a metric as experimental, stable, or deprecated in a way that tooling understands.
+* Ecosystem consumers (Thanos, Mimir, Grafana dashboards, alerting rules) have no authoritative reference for what a running Prometheus exposes.
+
+## Goals
+
+* [Required] Define Prometheus' internal telemetry as a formal OTel semantic convention registry (a single `registry.yaml`), making it the single machine-readable source of truth for every metric Prometheus exposes.
```
ArthurSens: Same here; it feels like you're tying the proposed solution to the goals. We want contract testing, better docs, etc etc and adopting OTel schemas looks like an awesome candidate, but adopting the schemas are not the goal per se.

proposals/0086-semconv-internal-telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+* [Required] Generate instrumentation code from the registry, eliminating hand-written metric definitions in Go.
+* [Required] Generate metric documentation from the registry that cannot drift from the implementation.
```
ArthurSens: Same thing here.

I'd rewrite the goals to something like "Eliminate handwritten metric descriptors" and "Auto-generated telemetry documentation".

proposals/0086-semconv-internal-telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+* [Nice to have] Enable contract testing through `promtool`: validate that a running Prometheus exposes exactly what the registry says, with no OTel Collector and no Weaver binary in the test path.
```
ArthurSens: why not use the tools that already exist for this? 🙃

proposals/0086-semconv-internal-telemetry.md
```diff
@@ -0,0 +1,255 @@
...
+The same Weaver invocation that produces `metrics.gen.go` also produces a `README.md` per package: a complete structured reference of every metric, including name, type, unit, label semantics, stability level, and examples. Because both files are rendered from the same `registry.yaml`, documentation cannot drift from the code.
+
+### Contract testing
+
+Contract testing checks a running Prometheus against the registry, with no OTel Collector and no Weaver binary in the test path.
+
+The primary surface is `promtool`. It already reads text exposition from stdin, already runs `promlint` over it, and already exits 3 when it finds problems:
+
+```
+curl -s http://localhost:9090/metrics | promtool check metrics --schema registry.yaml
```
ArthurSens: By default, Prometheus does not emit all metrics it is instrumented with. For example, if Remote Write isn't configured for an instance, it won't expose any Remote Write metrics.

What would happen in that situation, should `check metrics` fail because the metric exists in the schema but not in the /metrics endpoint?

ArthurSens: It also exposes other metrics that aren't Prometheus related, like go and http metrics, what happens in these situations?

ArthurSens (commented):
Had some minutes in the airport to review this, it's not a complete review though 😅

proposals/0086-semconv-internal-telemetry.md
nicolastakashi: Fair, rewrote them to lead with the outcome: one machine-readable description, no hand-written descriptors, docs that can't drift.

But in general adopting semconv is the proposal, title included, do you think we should switch and go on a direction that we want to explore alternatives for that or build a prometheus schema?

proposals/0086-semconv-internal-telemetry.md
nicolastakashi: addressed: https://github.com/prometheus/proposals/pull/86#discussion_r3897455752

proposals/0086-semconv-internal-telemetry.md
nicolastakashi: addressed: https://github.com/prometheus/proposals/pull/86#discussion_r3897455752

proposals/0086-semconv-internal-telemetry.md
nicolastakashi: live-check is an explicit alternative now instead of a buried paragraph. It wants a Collector in the test path and asserts against post-translation OTLP, so a translation bug reads as a registry violation. A Go test off Describe() runs where every other test already runs.

proposals/0086-semconv-internal-telemetry.md
nicolastakashi: Good catch. Cut the whole section and `--schema`, this is not related to the adoption of a registry anyway.

metalmatze:
Thanks for putting this together, @nicolastakashi.

I'm on board with the general direction. Generating metric definitions instead of hand-writing them is a good thing.

That said, I want to echo @roidelapluie on lighter built-in alternatives and @ArthurSens on tying the solution to the goals. For metrics alone, I'm not sure OTel semconv is the right fit, since a lot of what actually matters to us ends up in `annotations.prometheus`. A smaller Prometheus-specific project, or Weaver gaining a client_golang generator upstream, would get us most of the value.

The bigger point: if we pull Weaver in, we should standardize tracing and logging the same way. Metrics alone mostly buys us a bit more consistency, which I don't think is worth it on its own. All three signals would be.

Do you want to scope traces and logs here, or is there a reason metrics need to stand alone first?

aknuds1 (approved):
LGTM

PR #87, opened 2026-06-29
add detail about interaction with zero injection

ywwg:
as per the meeting notes

### DIFF
--- a/proposals/0060-st-storage.md
+++ b/proposals/0060-st-storage.md
@@ -23,7 +23,7 @@
   * [PROM-48 (Delta type)](https://github.com/prometheus/proposals/pull/48), [Delta WG](https://docs.google.com/document/d/1G0d_cLHkgrnWhXYG9oXEmjy2qp6GLSX2kxYiurLYUSQ/edit)
 
 > TL;DR: We propose to extend Prometheus TSDB storage sample definition to include an extra int64 that will represent the start timestamp (ST) (previously called created timestamp (CT)) for the cumulative types as well for the future delta temporality ([PROM-48](https://github.com/prometheus/proposals/pull/48)).
-> We propose introducing persisting ST logic behind a single flag `st-storage`. Once implemented, we could propose eventual removal of the `created-timestamps-zero-injection` experimental feature.
+> We propose introducing persisting ST logic behind a single flag `st-storage`.
 
 ## Why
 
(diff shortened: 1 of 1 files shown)


bwplotka (approved):
Thanks! We can also clarify perhaps what can we improve e.g. adding "stZeroInjection" to scrape config in future and removing FF flag (if you agee).

ywwg:
> Thanks! We can also clarify perhaps what can we improve e.g. adding "stZeroInjection" to scrape config in future and removing FF flag (if you agee).

done.  punted :)

proposals/0060-st-storage.md
```diff
@@ -203,6 +204,16 @@ Notably, given persistence of this feature, similar to example storage, if users
 
 This feature could be considered to be switched to opt-out, only after it's finished (this proposal is fully implemented) stable, provably adopted and when the previous LTS Prometheus version is compatible with this feature.
 
+#### Interaction with `created-timestamps-zero-injection`
+
+The existing zero-injection feature is useful and has been around long enough that some large users are relying on it. The new ST support will be in addition to or an alternative to zero-injection. In order to better differentiate zero-injection from st-per-sample, we propose to rename `created-timestamps-zero-injection` to `start-time-injection-at-scrape`.
```
bwplotka, line 209: I don't think we want to rename feature flag.

proposals/0060-st-storage.md
```diff
@@ -203,6 +204,16 @@ Notably, given persistence of this feature, similar to example storage, if users
...
+
+We also propose that the st setting be be tri-state on a per-series basis. Start Time per Sample support can be either `off`, `scrape`, or `per-sample`. Large installations may have a mix of implementations and there is no compatibility concern with having series have differing settings.
```
bwplotka, line 211: Are we sure we want per series? Can we configure it per series?

I'd say per scrape job is perhaps enough. 


> either `off`, `scrape`, or `per-sample`

Do we mean scrape job is planned to be either `off`, `stZeroInjection`, `stPerSample`?

proposals/0060-st-storage.md
```diff
@@ -203,6 +204,16 @@ Notably, given persistence of this feature, similar to example storage, if users
...
+
+In the case where a series has both scrape-injected zeroes and st-per-sample, the scrape-injected zeros take precedence.
```
bwplotka, line 213: Isn't above already blocking this (tri state)? 

Also you can't detect that after a fact and this statement assumes we will?

proposals/0060-st-storage.md
```diff
@@ -203,6 +204,16 @@ Notably, given persistence of this feature, similar to example storage, if users
...
+
+We also intend to graduate zero-injection from an experimental feature, which could involve replacing the existing flag with, for example, a `stZeroInjection` term to the scrape config. The exact implementation is out of scope of this proposal.
```
bwplotka, line 215: So why we explained before exactly how we planned it? (:

proposals/0060-st-storage.md
ywwg: This refers to when tsdb blocks being queried have both, for some reason. Ideally we shouldn't end up with data like that.

proposals/0060-st-storage.md
ywwg: People asked for per-series configuration, so that's what I wrote down, but I didn't want to specify exactly where that configuration would go.  You mention scrape job, other people seemed to agree with "per series"... that's why I didn't want to continue to slow down this PR with that detail and leave it for implementation

proposals/0060-st-storage.md
ywwg: ok, someone in the meeting mentioned renaming it so that's why I wrote it down

proposals/0060-st-storage.md
ywwg: I was hoping to add just enough text to indicate the intent, not the implementation, so we could merge the proposal

PR #89, opened 2026-07-01
Add query cost estimation and limits proposal

roidelapluie:


proposals/00089-query-cost.md
```diff
@@ -0,0 +1,133 @@
...
+The instant and range endpoints also gain a `cost` parameter. When set, the response `data` carries an estimated-vs-actual comparison:
+
+```json
+"cost": {
+  "estimated": { "seriesTouched": 42, "samplesScanned": 5040 },
+  "actual":    { "seriesTouched": 40, "samplesScanned": 4980, "peakSamples": 320 }
+}
+```
+
+Note: `cost=1` adds a second index lookup on top of executing the query.
```
charleskorn, line 86: nit: why `cost=1` and not `cost=true`? Do values of `cost` other than 0 and 1 have a meaning?

proposals/00089-query-cost.md
```diff
@@ -0,0 +1,133 @@
...
+**3. Limits.** Three reloadable knobs under `global:`:
+
+```yaml
+global:
+  query_max_series: 0            # 0 = no limit
+  query_max_samples_scanned: 0
+  query_max_duration: 0s
+```
+
+These are enforced *during* execution against the query's actual running cost, not against the estimate: a query is rejected as soon as it loads too many series or scans too many samples, and `query_max_duration` surfaces as a query timeout. A client may lower any ceiling for a single request via `max_series`, `max_samples_scanned`, `max_query_duration`; these can only tighten, never loosen, the operator-set value. The estimate is never used to reject a query — enforcement is always on the real cost.
```
charleskorn, line 97: If a request tries to override these limits to higher values than what is set on the server, is the request rejected? Or are the requested limits ignored?

proposals/00089-query-cost.md
```diff
@@ -0,0 +1,133 @@
...
+
+Note: `cost=1` adds a second index lookup on top of executing the query.
+
+**3. Limits.** Three reloadable knobs under `global:`:
+
+```yaml
+global:
+  query_max_series: 0            # 0 = no limit
+  query_max_samples_scanned: 0
+  query_max_duration: 0s
```
charleskorn, line 94: Is this different to the existing `-query.timeout` flag and `timeout` URL parameter?

proposals/00089-query-cost.md
```diff
@@ -0,0 +1,133 @@
...
+* Not a slow-query log.
+* Not per-tenant configuration, as Prometheus is not multi-tenant. Limits are global, with per-query lowering only.
+* Not exact cost prediction. The estimate is intentionally cheap and approximate.
+
+## How
+
+Three pieces, all gated by `--enable-feature=query-cost`.
+
+**1. Estimation (`promql.EstimateCost`).** Parse the query, walk it for every vector and matrix selector, compute the effective time window each selector reads (mirroring the engine's `getTimeRangesForSelector`/`populateSeries`), and ask storage for the series count per selector via a single querier over the union window. `SeriesTouched` is the sum across selectors — an upper bound, because a series shared between selectors is counted once per selector. `SamplesScanned` models the engine's incremental per-step reads (full range window at step 0, then only the samples that advance past the previous cutoff), scaled by a measured average per-point cost so native-histogram points are sized by bucket rather than counted as one float unit. The estimate is index-only apart from decoding at most `histogramSampleLimit` (50) points per selector to size histograms.
```
charleskorn, line 57: Can you elaborate on how the sample count estimation would work? Would this require decoding each chunk, or would it use the sample count stored on each chunk to avoid decoding chunks?

charleskorn, line 57: If I'm understanding correctly, the goal is to roughly estimate the value of the "samples read" statistic introduced in https://github.com/prometheus/prometheus/pull/18081 (rather than the "total samples" statistic that existed before that PR). Is that correct? If so, it might be worth mentioning that here.

proposals/00089-query-cost.md
saswatamcode, line 97: I wonder if it might actually make sense to enforce the limit based on the estimation, rather than allowing the query to run up until a limit is hit.

Feels wasteful to let a query that will likely be limited, run and fetch data.

proposals/00089-query-cost.md
roidelapluie: Maybe we can have cost=2 later that e.g. read chunks metadata ?

proposals/00089-query-cost.md
roidelapluie: "A client may lower any ceiling". The request is not rejected but you will be capped at the server's limit.

proposals/00089-query-cost.md
roidelapluie: It is not, more a normalization of it.

proposals/00089-query-cost.md
roidelapluie: There are 2 options.

In Prometheus we can take the scrape interval for estimationm,
or we can estimate based on the number of chunks and do x120.

Looking at the chunks header would be almost as I/O expensive as running the query.

proposals/00089-query-cost.md
roidelapluie: I'd say it will depend on how we can accurately estimate the limit upfront.

charleskorn: My 2c is that requests which ask for a higher value should be rejected - this makes it very clear that the requested limit wasn't applied.

proposals/00089-query-cost.md
charleskorn: Will the existing flag / parameter be deprecated? Otherwise we have two ways to do the same thing.

proposals/00089-query-cost.md
charleskorn: >In Prometheus we can take the scrape interval for estimationm,

Would this apply to chunks written with remote write? (what scrape interval would be used?)

roidelapluie: Yes if we go that way.

I am exploring sampling with a limited number of chunks as well.

lamida, line 57: > (50) points per selector 

Any reason on picking 50?

lamida, line 57: Nits. Should we use canonical term of instant vector and range vector selector instead of internal struct naming vector and matrix selector?

```suggestion
**1. Estimation (`promql.EstimateCost`).** Parse the query, walk it for every instant vector and range vector selector, compute the effective time window each selector reads (mirroring the engine's `getTimeRangesForSelector`/`populateSeries`), and ask storage for the series count per selector via a single querier over the union window. `SeriesTouched` is the sum across selectors — an upper bound, because a series shared between selectors is counted once per selector. `SamplesScanned` models the engine's incremental per-step reads (full range window at step 0, then only the samples that advance past the previous cutoff), scaled by a measured average per-point cost so native-histogram points are sized by bucket rather than counted as one float unit. The estimate is index-only apart from decoding at most `histogramSampleLimit` (50) points per selector to size histograms.
```

proposals/00089-query-cost.md
```diff
@@ -0,0 +1,133 @@
...
+* `--query.max-samples` caps peak samples in memory, not the total scanned.
+* `--query.timeout` and `--query.max-concurrency` are process-wide flags, not reloadable and not per-query.
+* Nothing tells a user, an autocomplete UI, or an alerting rule author how heavy a query is *before* it runs.
+
+Operators want ceilings they can tune without a restart. Users and tools (Grafana, dashboards, recording rules) want a cheap way to gauge cost up front so they can refuse or rewrite a query before it lands on the server.
+
+### Pitfalls of the current solution
+
+* The existing limits are set at startup. Changing them means a restart.
+* They are global. A single tenant or dashboard cannot be given a tighter budget.
```
lamida, line 28: Nits. Mentioned in the non-goal below, but isn't we know prometheus doesn't support multiple tenant?

proposals/00089-query-cost.md
```diff
@@ -0,0 +1,133 @@
...
+* There is no pre-execution estimate. The only way to learn a query's cost today is to run it, which is exactly what we want to avoid for the expensive ones.
+* `--query.max-samples` measures peak in-memory samples, which does not map cleanly to "how much index and how many samples did this touch".
+
+## Goals
+
+* Give a cheap, index-based cost *estimate* (series touched, samples scanned) without executing the query.
```
lamida, line 34: Should we clear up what is executing the query here? Is this referring to expanding posting or decoding chunk? Maybe only me, we can keep this as it is if this is considered clear.

proposals/00089-query-cost.md
roidelapluie: I have enforced cost=true

proposals/00089-query-cost.md
roidelapluie: Will make that change.

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,133 @@
...
+
+Note: `cost=1` adds a second index lookup on top of executing the query.
+
+**3. Limits.** Three reloadable knobs under `global:`:
+
+```yaml
+global:
+  query_max_series: 0            # 0 = no limit
+  query_max_samples_scanned: 0
+  query_max_duration: 0s
```
roidelapluie: I have deprecated them

proposals/00089-query-cost.md
roidelapluie: I propose to use ChunksQuerier instead

roidelapluie: Yes, I have mentioned 18081 explicitly now.

roidelapluie: This is arbitrary, I think it could work.

proposals/00089-query-cost.md
roidelapluie: It makes it clear that we do not tie that to prometheus exporter toolkit authentication layer

proposals/00089-query-cost.md
roidelapluie: I am fine either way. fine as is.

proposals/00089-query-cost.md
roidelapluie: I removed index-based. This is the overal goals.

roidelapluie:
Hello!

I made substantial changes to the proposal and I am working on the Proof of Concept.

cc @charleskorn @lamida

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,141 @@
...
+### Audience
+
+Operators running shared Prometheus servers, and UI/tooling authors (Grafana) that build queries on a user's behalf.
+
+## Non-Goals
+
+* Not replacing `--query.max-samples`, `--query.timeout`, or `--query.max-concurrency`.
+* Not a billing or chargeback system. The numbers are upper bounds, not exact accounting.
+* Not a slow-query log.
+* Not per-tenant configuration, as Prometheus is not multi-tenant. Limits are global, with per-query lowering only.
```
zhehao-grafana, line 50: I wonder whether these configurations can be integrated with a RBAC system so the limit can be applied at either user or team level

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,141 @@
...
+* API tests for the new endpoints and the `cost` parameter.
+* OpenAPI golden files updated for the new paths and schemas.
+
+### Migration
+
+Purely additive and behind a feature flag. Default config (all limits `0`) changes no behaviour.
+
+### Known unknowns
+
+* **Estimate accuracy.** `SeriesTouched` still over-counts shared series and series with no in-window samples; `SamplesScanned` assumes samples land exactly at the measured or scrape interval and that sampled series are representative. Partially resolved when a selector's real window fits within the 50-chunk/50-series sample budget, the measurement is now taken from that real window instead of extrapolated from a nearby proxy window, so small selectors get an exact rather than approximate density (see How, section 1). Larger selectors still extrapolate from a bounded sample. Is an upper bound the right contract for those, or do we want something tighter?
```
zhehao-grafana: over-counting is fine and we can check the % between estimate and actual count after accumulating the number of real queries

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,141 @@
...
+* **Config surface.** Should limits live under `global:`, or a dedicated `query:` section?
+* **Units** Is it enough to return number of samples/series or do we want to return bytes?
+
+## Alternatives
+
+1. **Estimate from postings cardinality directly, bypassing `storage.Querier`.** Cheaper, but ties the estimator to the TSDB index and breaks for any other `storage.Queryable` (remote read, federation). Using the portable `Select` path keeps it storage-agnostic.
+2. **Reject queries based on the estimate.** Rejected as the default: the estimate is an upper bound and can be wrong in both directions, so rejecting on it would refuse queries that would actually run fine. Enforcement is on real cost; the estimate is advisory only. There is a fair argument that letting a query that will almost certainly be limited run and fetch data anyway is wasteful. If the estimate proves accurate enough in practice (validated via the `cost` object's estimated-vs-actual comparison), an *opt-in* upfront rejection — reject before execution when the estimate clearly exceeds a ceiling — could be added later as a follow-up without changing the real-cost enforcement that remains the backstop.
```
zhehao-grafana: as long as the difference between estimate and actual is too high (e.g. more than 20%), users will ask for a way to reject the query before it runs

proposals/0089-query-cost.md
roidelapluie: This could be done in Mimir, but for Prometheus downstream there is nothing around RBAC.

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,141 @@
...
+* **Chunk sampling (always-on, not opt-in).** The estimator samples automatically, with no user-facing knob, whenever the storage exposes `storage.ChunkQueryable`: it reads up to a fixed `chunkSampleLimit` (50) chunks' `NumSamples` header to measure the selector's real sample interval, and decodes the first point of up to a fixed `histogramSampleLimit` (50) series to size native-histogram points by bucket count. Reading a chunk header is far cheaper than decoding its samples, so this stays much cheaper than executing the query.
+
+Sampling prefers the *real* query window over a nearby proxy window whenever that real window is cheap enough: if a selector's actual chunk count (for density) or series count (for point cost) already fits within the 50-item budget, the estimator samples directly from `[sel.mint, sel.maxt]` and gets an exact rather than extrapolated measurement. Only when the real window has more chunks/series than the budget affords does it fall back to sampling a bounded, narrow window near the query's end and extrapolating.
```
krajorama: What's proxy window? It's not defined anywhere.

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,141 @@
...
+```json
+{
+  "estimate": {
+    "seriesTouched": 42,
+    "samplesScanned": 5040
+  }
+}
+```
+
+The instant and range endpoints also gain a `cost=true` boolean parameter. When set, the response `data` carries an estimated-vs-actual comparison:
```
krajorama: Shouldn't `cost` be an enum? Meaning:  omitted, or `none`, or `actual` or `compare`. Or something like that? The current solution implies we're re-running the cost estimation even if it was estimated before up front.

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Migration
+
+Purely additive and behind a feature flag. Default config (all limits `0`) changes no behaviour.
+
+### Known unknowns
+
+* **Estimate accuracy.** `SeriesTouched` still over-counts shared series and series with no in-window samples; `SamplesScanned` assumes samples land exactly at the measured or scrape interval and that sampled series are representative. Partially resolved when a selector's real window fits within the 50-chunk/50-series sample budget, the measurement is now taken from that real window instead of extrapolated from a nearby proxy window, so small selectors get an exact rather than approximate density (see How, section 1). Larger selectors still extrapolate from a bounded sample. Is an upper bound the right contract for those, or do we want something tighter?
+* **Scrape interval.** Mostly resolved for TSDB-backed storage: the estimator measures the real density from chunk metadata automatically whenever it's available, with no configuration needed. The caller-supplied scrape interval remains a fallback only for a plain `storage.Queryable` with no chunk metadata (e.g. some remote-read backends), or when a selector's window has nothing to sample.
```
krajorama: You say "caller-supplied" scrape interval, but there was no mention of this in the API, only `cost` parameter.

Also note that if we had native metadata and we propagated it (lot o ifs) , you might get accurate scrape interval.

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,141 @@
...
+
+The per-point density can be derived two ways:
+
+* **Scrape-interval (fallback, index-only).** Assume samples land at the global scrape interval and compute window ÷ interval. Cheapest, but wrong for series scraped at a different interval and for remote-written series, which have no scrape interval at all. Used only when nothing can be sampled (see below).
+* **Chunk sampling (always-on, not opt-in).** The estimator samples automatically, with no user-facing knob, whenever the storage exposes `storage.ChunkQueryable`: it reads up to a fixed `chunkSampleLimit` (50) chunks' `NumSamples` header to measure the selector's real sample interval, and decodes the first point of up to a fixed `histogramSampleLimit` (50) series to size native-histogram points by bucket count. Reading a chunk header is far cheaper than decoding its samples, so this stays much cheaper than executing the query.
```
krajorama: Does it do this reading of 50 chunks+50 samples for every series from SeriesTouched?

roidelapluie: No. Budgets are per selector, not per series. SeriesTouched itself is index-only. On top of that, per selector: at most 50 chunks across all its series, and at most 50 series decoded (first point only) to size native-histogram points. A selector matching a million series still costs 50 chunk headers plus 50 first points.

proposals/0089-query-cost.md
roidelapluie: You are right, it is not an API parameter. It is `global.scrape_interval` from the config, passed to a `promql.EstimateCost` by the API layer. Clarified.

Agreed with metadata, added a note.

proposals/0089-query-cost.md
roidelapluie: `cost=true` re-runs the estimation after execution, it is not free. But an actual-only mode would duplicate stats because samplesRead, totalSeries and peakSamples already come from qry.Stats(). `cost` only adds the estimated side and the pairing, so a boolean is enough for now.

proposals/0089-query-cost.md
roidelapluie: It's fallback sampling window: `max(5m, 8 × scrape interval)`, capped at 30m, ending at the selector's end. Updating the doc.

proposals/0089-query-cost.md
roidelapluie: Yes, I proposed that as a follow up in the doc, thanks.

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,172 @@
...
+
+`SeriesTouched` estimates series reads summed across selectors. Repeated selectors may count a series more than once, as execution also can. It is not a distinct-series count. Index entries may lack in-window samples, and expressions such as `info()` can select additional series at runtime. When such selections cannot be estimated without evaluation, the response should warn that the estimate is incomplete. Neither this figure nor the sample estimate is a guaranteed upper bound.
+
+`SamplesScanned` should approximate the engine's `samplesRead` statistic. Range queries should account for incremental reads as evaluation advances, rather than counting every overlapping range window in full. Disjoint windows, nested evaluation grids and expressions evaluated only once must be treated according to their execution semantics.
```
DanCech: Any reason not to call this `samplesRead` also? presumably the actual value in the cost output would be the `samplesRead` number?

roidelapluie: Yes, agreed. The actual value comes directly from samplesRead in my PoC. I will use the same name to avoid unneeded confusion.

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,172 @@
...
+  "cost": {
+    "estimated": { "seriesTouched": 42, "samplesRead": 5040 },
+    "actual":    { "seriesTouched": 40, "samplesRead": 4980, "peakSamples": 320 }
+  }
+}
+```
+
+`cost` is a boolean on/off switch, not a set of levels. It runs a fresh estimation after execution, so it adds work and can cost more than executing a cheap query. If that estimation fails, the successful query result is preserved, `cost` is omitted, and a warning explains the failure.
+
+Actual counters should also be available without another estimation through `stats=true`: `data.stats.samples.seriesTouched`, `samplesRead`, and `peakSamples`. This lets calibration tools call an estimate endpoint and then execute with `stats=true`, without triggering a second estimate through `cost=true`.
```
bwplotka, line 107: so cost=true and stats=true would give:

```
{
  "cost": {
    "estimated": { "seriesTouched": 42, "samplesRead": 5040 },
    "actual":    { "seriesTouched": 40, "samplesRead": 4980, "peakSamples": 320 }
  }
  "stats" {  { "seriesTouched": 40, "samplesRead": 4980, "peakSamples": 320  ....}
}
```

Shouldn't cost do the estimation and stats remain for "actual" data?

proposals/0089-query-cost.md
```diff
@@ -0,0 +1,172 @@
...
+**3. Limits.** Add three reloadable settings under `global:`:
+
+```yaml
+global:
+  query_max_series: 0            # 0 = no limit
+  query_max_samples_read: 0
+  query_max_duration: 0s        # 0s = fall back to --query.timeout
+```
+
+Enforce these limits *during* execution against actual running cost, not the estimate. The budget must cover the entire query, including input consumed by nested subqueries and expressions evaluated only once. Series and sample limit violations return a `cost_limit` API error; duration limits surface as timeouts. These counters constrain evaluator consumption, but do not prevent all index enumeration or buffering that storage may perform before yielding data.
```
bwplotka, line 122: Wouldn't it be more clear if we call those `_estimated_`.. 🤔? So we know which one is which?

bwplotka (approved):
I am supportive, this looks like a useful step forward to get cheaper cost estimate.

DanCech (approved):

proposals/0089-query-cost.md
DanCech: are you referring to the limit settings? If I'm reading this correctly the limits would be enforced during execution, not by pre-running the estimate and blocking before executing.

proposals/0089-query-cost.md
DanCech: it is somewhat duplicative, but I guess we also need to think about what happens when the user asks for cost=true without stats=true

roidelapluie: I believe that cost and stats could differ in the future, if we want to take into account e.g. function prices etc.

proposals/0089-query-cost.md
roidelapluie: yes, those are not estimates

proposals/0089-query-cost.md
bwplotka: I don't buy the difference argument - the whole point is that cost can be done in separate call and stats give you exact number to compare with so those has to stay similar, right? Why duplicate?

Can cost=true give you no actual data and you need stats true OR cost true gives you relevant actual stats in stats 🤔

A bit odd

roidelapluie: I expect we may add more fields to “cost” or change how its value is calculated as we refine how we account for metrics, as we move on the actual cost of a query. Keeping this extra experimental field would give us that flexibility.

bwplotka: I do buy the argument that we cannot change stats much. Ok, let's keep this duplication

PR #94, opened 2026-09-11
proposal: Allow unquoted Unicode letters and dots in metric and label names

roidelapluie:


juliusv:
Sounds good to me, thanks :+1: 

I don't know how annoying it will be to implement this behind a feature flag, especially since case differentiation in lexing/parsing (both backend + frontend) could become cumbersome. If that turns out to be too much trouble, maybe we can even relax the requirement that it needs to be behind a feature flag.

Btw. I started a vibe-coded implementation in https://github.com/prometheus/prometheus/tree/juliusv-vibe/promql-unquoted-dots a few weeks back just to explore things, see commit https://github.com/prometheus/prometheus/commit/87e9695002378200c17508b2308f35e918842a4e. That still doesn't cover everything (e.g. printing in the backend or frontend), but feel free to ignore that branch or use it as an inspiration (or copy whatever you like).

Maybe one thing to note that popped up in that session: the frontend's lezer parsing system does not have an equivalent of `unicode.IsLetter`, so this is the solution that Claude came up with:

```
  // Besides ASCII letters, names may contain any non-ASCII character that is a
  // letter. Since the tokenizer cannot check Unicode character classes, any
  // non-ASCII character is accepted here. Names containing non-letters still
  // have to be quoted, which the PromQL parser in Prometheus enforces.
  // The ranges below cover all non-ASCII characters except whitespace.
  nameLetter {
    std.asciiLetter |
    $[\u{a1}-\u{167f}\u{1681}-\u{1fff}\u{200b}-\u{2027}\u{202a}-\u{202e}\u{2030}-\u{205e}\u{2060}-\u{2fff}\u{3001}-\u{10ffff}]
  }
  Identifier { (nameLetter | "_" | ":") (nameLetter | std.digit | "_" | ":" | ".")* }
  LabelName { (nameLetter | "_") (nameLetter | std.digit | "_" | ".")* }
  ```
  
That sounds like it would allow the user to input characters that parse fine on the frontend but cause an error on the backend. Would be nice if there was a better solution, but I haven't looked into it myself yet.

roidelapluie:
https://github.com/prometheus/prometheus/pull/19683

PR #97, opened 2026-09-22
PROM-0097: PromQL Metadata querying / promotion.

roidelapluie:
This proposal was writting by hand. Grammar fixed by AI.

I asked AI 5 examples queries:

```promql
up{target[deployment.environment.name]="production"} == 0
```

```promql
up{target[team.name] as team} == 0
```

```promql
queue_depth{resource[deployment.environment.name]="production" as environment} > 1000
```

```promql
sum by (service) (
  rate(http_requests_total{resource[service.name] as service}[5m])
)
```

```promql
sum by (cost_center) (
  process_resident_memory_bytes{resource["billing/cost-center"] as cost_center}
)
```


proposals/0097-promoting-promql-metadata.md
```diff
@@ -0,0 +1,95 @@
...
+
+2. Native metadata have a namespace.
+
+Native metadata have namespaces or scopes, such as resource or target.
+
+3. Native metadata are referenced as `namespace[cpu.foo]` or `namespace["super+name"]` (quoting follows the rules for label and metric names).
+
+4. Native metadata can optionally be promoted to labels with the `as` keyword.
+
+`foo{resource[power.status] as power_status}` `foo{namespace["super+name"] as "supername"}`
```
bwplotka, line 61: So what are the keywords allowed here? `resource` `namespace`? What else?

roidelapluie: This proposal is agnostic of the keywords. The keywords are actual namespaces of metadata and should be defined in a broader metadata design doc like the ones that exist. I can think of resource for OTLP: resource, k8s, host, service...

bwplotka: Thanks. I am missing an exact definition of "namespace" or "scope", especially how it would be reflected in Prometheus world. Also if we expect UTF-8 in those (:

Maybe it's out of scope, but then how we could understand limitations of `<prefix>[<suffix>]` syntax?

If we mean here OTel entity [namespaces](https://opentelemetry.io/docs/specs/semconv/registry/entities/) and let's take [`process.executable`](https://opentelemetry.io/docs/specs/semconv/registry/entities/process/#process-executable))

Then should I go for

`cpu_info{process["executable.name"]="docker" as exec}`
or 
`cpu_info{"process.executable"[name]="docker" as exec}`

..and if this "namespacing" is required, why not `cpu_info{["process.executable.name"]="docker" as exec}`

Then can we guarentee some namespace is always there for metadata?

roidelapluie: I believe we can have better and will do a second proposal later.

